PR #20, opened 2019-01-25
DDL dialect functions

### LINKED ISSUE #11: DDL dialect functions
override and implement as needed, including but not limited to: 

canCreateCatalog - default works
getCreateCatalogCommand - default works
getDropCatalogCommand - default works
canCreateSchema - override
getCreateSchemaCommand - override
getDropSchemaCommand - override
getCurrentSchemaCommand - override
getSchemaNameResolver - override
hasAlterTable - default works
dropConstraints - override
qualifyIndexName - override
getAddColumnString - override
getAddColumnSuffixString - default works
getDropForeignKeyString - override
getTableTypeString - default works
getAddForeignKeyConstraintString - override
getAddPrimaryKeyConstraintString - override
hasSelfReferentialForeignKeyBug - default works
getNullColumnString  - default works
supportsCommentOn - default works
getTableComment - default works
getColumnComment - default works
supportsIfExistsBeforeTableName - default works
supportsIfExistsAfterTableName - default works
supportsIfExistsBeforeConstraintName - default works
supportsIfExistsAfterConstraintName - default works 
supportsIfExistsAfterAlterTable - default works
getDropTableString - default works

ChengyuanZhao:
fixes #11
overrides and tests the DDL dialect meta functions

### DIFF
--- a/src/main/java/com/google/cloud/spanner/hibernate/SpannerDialect.java
+++ b/src/main/java/com/google/cloud/spanner/hibernate/SpannerDialect.java
@@ -19,6 +19,7 @@
 package com.google.cloud.spanner.hibernate;
 
 import org.hibernate.dialect.Dialect;
+import org.hibernate.engine.jdbc.env.spi.SchemaNameResolver;
 import org.hibernate.mapping.Table;
 import org.hibernate.tool.schema.spi.Exporter;
 
@@ -36,6 +37,8 @@ public class SpannerDialect extends Dialect {
   public Exporter<Table> getTableExporter() {
     return this.spannerTableExporter;
   }
+
+  /* SELECT-related functions */
 
   @Override
   public boolean supportsCurrentTimestampSelection() {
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/cloud/spanner/hibernate/SpannerDialect.java
```diff
@@ -66,4 +69,79 @@ public boolean supportsUnionAll() {
...
+
+  @Override
+  public String[] getDropSchemaCommand(String schemaName) {
+    throw new UnsupportedOperationException(
+        "No drop schema syntax supported by " + getClass().getName());
+  }
+
+  @Override
+  public String getCurrentSchemaCommand() {
+    throw new UnsupportedOperationException("No drop schema syntax supported by "
```
meltsufin:
```suggestion
    throw new UnsupportedOperationException("No current schema syntax supported by "
```

src/main/java/com/google/cloud/spanner/hibernate/SpannerDialect.java
```diff
@@ -66,4 +69,79 @@ public boolean supportsUnionAll() {
...
+
+  @Override
+  public SchemaNameResolver getSchemaNameResolver() {
+    throw new UnsupportedOperationException(
+        "No schema name resolver supported by " + getClass().getName());
+  }
+
+  @Override
+  public boolean dropConstraints() {
+    return false;
```
meltsufin: Would "DROP INDEX" apply, since index provides a uniqueness constraint?

ChengyuanZhao: I don't think so here actually. The reason is Hibernate considers indexes and constraints two separate things and has a baked-in assumption that the UNIQUE word is associated with the "CONSTRAINT" word instead of the "INDEX" word (and has separate functions for indexes and constraints for us to override). What I'm getting at is that I think we need to say we don't support constraints and instead we need to drop the index when dropping table transparently to Hibernate. I have a feeling if we start crossing index and constraint wires with Hibernate it'll bite us later on.

meltsufin: Right, I think that in a typical RDMBS the indexes are dropped automatically when you do `DROP TABLE`. Does Spanner not do that? Does it require you to drop indexes separately?

ChengyuanZhao: Yes, in Spanner you must do this with separate statements.

ChengyuanZhao: Yea that's right. https://cloud.google.com/spanner/docs/data-definition-language#drop_table look at the note in blue.

meltsufin: interesting

PR #23, opened 2019-01-29
Setup testing queries SQL and DML

### LINKED ISSUE #22: Set up SELECT and DML queries tests based on TestEntity like we have for DDL

ChengyuanZhao:
fixes #22
sets up tests and provides example tests for generating complete SQL and DML

### DIFF
--- a/src/test/java/com/google/cloud/spanner/hibernate/GeneratedSelectStatementsTests.java
+++ b/src/test/java/com/google/cloud/spanner/hibernate/GeneratedSelectStatementsTests.java
@@ -1,0 +1,87 @@
+/*
+ * Copyright 2019 Google LLC
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA
+ */
+
+package com.google.cloud.spanner.hibernate;
+
+import com.google.cloud.spanner.hibernate.util.TestEntity;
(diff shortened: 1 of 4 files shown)


src/test/java/com/google/cloud/spanner/hibernate/GeneratedSelectStatementsTests.java
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  @Test
+  public void deleteDmlTest() {
+    openSessionAndDo(x -> x.createQuery("delete TestEntity where boolVal = true").executeUpdate());
+
+    List<String> statements = this.jdbcMockObjectFactory.getMockConnection()
+        .getPreparedStatementResultSetHandler().getPreparedStatements().stream()
+        .map(MockPreparedStatement::getSQL).collect(
+            Collectors.toList());
```
dmitry-s: looks like it should be a method to avoid code duplication

src/test/java/com/google/cloud/spanner/hibernate/GeneratedSelectStatementsTests.java
```diff
@@ -0,0 +1,114 @@
...
+
+    assertEquals("delete from test_table where boolVal=TRUE", statements.get(0));
+  }
+
+  @Test
+  public void selectJoinTest() {
```
meltsufin: There's still some code duplication here. Can we have a utility method like `testTranslation(String hsql, String spannerQuery)`?

src/test/java/com/google/cloud/spanner/hibernate/GeneratedSelectStatementsTests.java
ChengyuanZhao: done! refactored.

src/test/java/com/google/cloud/spanner/hibernate/GeneratedSelectStatementsTests.java
```diff
@@ -0,0 +1,115 @@
...
+        .build();
+    this.metadata =
+        new MetadataSources(this.registry).addAnnotatedClass(TestEntity.class)
+            .addAnnotatedClass(SubTestEntity.class).buildMetadata();
+  }
+
+  @Test
+  public void deleteDmlTest() {
+    testStatementTranslation(
+        x -> x.createQuery("delete TestEntity where boolVal = true").executeUpdate(),
```
meltsufin: Assuming that we'll have more than one update test, you can make it even cleaner by writing a utility method
`testUpdateStatementTranslation("delete TestEntity where boolVal = true", "delete from test_table where boolVal=TRUE")`

ChengyuanZhao: Yes , good point. will do.

PR #25, opened 2019-01-30
lock acquisition dialect functions

### LINKED ISSUE #7: Row Lockout dialect functions
Spanner doesn't have row lockouts. 

Override and implement as necessary including but not limited to:

supportsLockTimeouts
isLockTimeoutParameterized
getLockingStrategy
getForUpdateString
getWriteLockString
getReadLockString
forUpdateOfColumns
supportsOuterJoinForUpdate
getForUpdateString
getForUpdateNowaitString
getForUpdateSkipLockedString
appendLockHint
applyLocksToSql

ChengyuanZhao:
fixes #7
overrides functions related to lock acquisition

### DIFF
--- a/src/main/java/com/google/cloud/spanner/hibernate/SpannerDialect.java
+++ b/src/main/java/com/google/cloud/spanner/hibernate/SpannerDialect.java
@@ -18,9 +18,18 @@
 
 package com.google.cloud.spanner.hibernate;
 
+import java.io.Serializable;
+import java.util.Map;
+import org.hibernate.LockMode;
+import org.hibernate.LockOptions;
+import org.hibernate.StaleObjectStateException;
 import org.hibernate.dialect.Dialect;
+import org.hibernate.dialect.lock.LockingStrategy;
+import org.hibernate.dialect.lock.LockingStrategyException;
 import org.hibernate.engine.jdbc.env.spi.SchemaNameResolver;
+import org.hibernate.engine.spi.SharedSessionContractImplementor;
 import org.hibernate.mapping.Table;
+import org.hibernate.persister.entity.Lockable;
 import org.hibernate.tool.schema.spi.Exporter;
 
 /**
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/cloud/spanner/hibernate/SpannerDialect.java
```diff
@@ -144,4 +155,115 @@ public String getAddForeignKeyConstraintString(
...
+    return sql;
+  }
+
+  /**
+   * A locking strategy for the Cloud Spanner dialect that does nothing. Cloud Spanner does not
+   * support locking.
+   *
+   * @author Chengyuan Zhao
+   */
+  public static class DoNothingLockingStrategy implements LockingStrategy {
```
dzou: Remove the `public` modifier?

ChengyuanZhao: done. I made it package private (so the test can see it)

src/main/java/com/google/cloud/spanner/hibernate/SpannerDialect.java
```diff
@@ -144,4 +155,115 @@ public String getAddForeignKeyConstraintString(
...
+
+  @Override
+  public String applyLocksToSql(String sql, LockOptions aliasedLockOptions,
+      Map<String, String[]> keyColumnNames) {
+    return sql;
+  }
+
+  /**
+   * A locking strategy for the Cloud Spanner dialect that does nothing. Cloud Spanner does not
+   * support locking.
```
meltsufin: How does transaction-level locking fit into this?
https://cloud.google.com/spanner/docs/transactions

ChengyuanZhao: The locking here is at the query-level (part of SELECT queries) while the transaction locking setting is at the session level (you use a session to create many queries). The transaction locking and such isn't in the dialect from my examination and is handled by the driver implementation instead.

Issue #27, opened 2019-01-31
TableExporter.getSqlCreateStrings fails with NPE


meltsufin:
Many of the documentation tests seem to be failing with an NPE in `TableExporter.getSqlCreateStrings`.

```
Caused by: java.lang.NullPointerException
	at com.google.cloud.spanner.hibernate.SpannerTableExporter.getSqlCreateStrings(SpannerTableExporter.java:56)
	at com.google.cloud.spanner.hibernate.SpannerTableExporter.getSqlCreateStrings(SpannerTableExporter.java:35)
	at org.hibernate.tool.schema.internal.SchemaCreatorImpl.createFromMetadata(SchemaCreatorImpl.java:316)
	at org.hibernate.tool.schema.internal.SchemaCreatorImpl.performCreation(SchemaCreatorImpl.java:166)
	at org.hibernate.tool.schema.internal.SchemaCreatorImpl.doCreation(SchemaCreatorImpl.java:135)
	at org.hibernate.tool.schema.internal.SchemaCreatorImpl.doCreation(SchemaCreatorImpl.java:121)
	at org.hibernate.tool.schema.spi.SchemaManagementToolCoordinator.performDatabaseAction(SchemaManagementToolCoordinator.java:155)
	at org.hibernate.tool.schema.spi.SchemaManagementToolCoordinator.process(SchemaManagementToolCoordinator.java:72)
	at org.hibernate.internal.SessionFactoryImpl.<init>(SessionFactoryImpl.java:309)
	at org.hibernate.boot.internal.SessionFactoryBuilderImpl.build(SessionFactoryBuilderImpl.java:462)
	at org.hibernate.jpa.boot.internal.EntityManagerFactoryBuilderImpl.build(EntityManagerFactoryBuilderImpl.java:938)
	... 13 more
```

ChengyuanZhao:
@meltsufin which set of tests does this one fall under (the --tests <test_name>)? I think we would be throwing an UnsupportedOperationException here instead, but want to be sure..

meltsufin:
Sure, try this one:
```
python3 run-tests.py SQLTest.test_sql_jpa_all_columns_scalar_query_example
```

ChengyuanZhao:
closed by #28

PR #28, opened 2019-02-01
Enforce PK required check

ChengyuanZhao:
Check that a property annotated as the ID is present in all versions of Hibernate including ones where no-ID entities are allowed.

### DIFF
--- a/src/main/java/com/google/cloud/spanner/hibernate/SpannerTableExporter.java
+++ b/src/main/java/com/google/cloud/spanner/hibernate/SpannerTableExporter.java
@@ -50,6 +50,12 @@ public class SpannerTableExporter implements Exporter<Table> {
     /* The current implementation does not support interleaved tables for collections
      * or UNIQUE constraints/indexes for relationships
      * */
+
+    if (!table.hasPrimaryKey()) {
+      throw new UnsupportedOperationException("Cloud Spanner requires tables and entities to have"
+          + " at least one ID column to act as the Primary Key.");
+    }
+
     String createTableTemplate = this.spannerDialect.getCreateTableString()
         + " {0} ({1}) PRIMARY KEY ({2})";
 


meltsufin:
Does this fix #27?

ChengyuanZhao:
@meltsufin confirmed that this now replaces the NPE with the new UnsupportedOperationException that I just added. @dzou

dzou:
Gotcha. What exactly was the root problem here? Was the hibernate test creating entities that do not have a PK?

ChengyuanZhao:
@dzou Yes that's right. The entity metadata the test provides has .hasPrimaryKey() == false.

meltsufin:
Can we align the tests with the actual version of hibernate we're targeting and make sure the test passes then?

src/main/java/com/google/cloud/spanner/hibernate/SpannerTableExporter.java
```diff
@@ -50,6 +50,12 @@ public SpannerTableExporter(SpannerDialect spannerDialect) {
     /* The current implementation does not support interleaved tables for collections
      * or UNIQUE constraints/indexes for relationships
      * */
+
+    if (!table.hasPrimaryKey()) {
+      throw new UnsupportedOperationException("Cloud Spanner requires tables and entities to have"
```
meltsufin: I think this is more of an `IllegalArgumentException`. UOE is usually for completely unimplemented methods.

ChengyuanZhao: I think that case is better covered by the `UnimplementedException`. Here I think it really is an unsupported operation because Hibernate assumes not having a PK is a valid thing to create because standard SQL DBs accept it and it is generally a valid operation to create such a table (vs. generally illegal), but we cannot support it.

PR #30, opened 2019-02-19
use quoted names

dmitry-s:


src/test/java/com/google/cloud/spanner/hibernate/util/TestEntity.java
```diff
@@ -31,7 +31,7 @@
  * @author Chengyuan Zhao
  */
 @Entity
-@Table(name = "test_table")
+@Table(name = "`test_table`")
```
meltsufin: Is this required in the annotation? It looks a bit ugly this way an unconventional.

dmitry-s: That's not required, but we need to support that. There are some tests in hibernate that use it.

PR #31, opened 2019-02-26
Add no-op exporters for the Dialect

### LINKED ISSUE #9: Non-functional/Unsupported features Exporters
This involves providing implementations that generate BLANK statements for:

sequences
foreign key constraint
unique constraint
auxiliary constraint

dzou:
This adds Exporters which return empty string[] for unsupported features of the dialect.

Fixes #9.

### DIFF
--- a/src/main/java/com/google/cloud/spanner/hibernate/SpannerDialect.java
+++ b/src/main/java/com/google/cloud/spanner/hibernate/SpannerDialect.java
@@ -50,9 +50,7 @@ public class SpannerDialect extends Dialect {
 
   private static final LockingStrategy LOCKING_STRATEGY = new DoNothingLockingStrategy();
 
-  private static final Exporter<Sequence> NOOP_SEQUENCE_EXPORTER = new EmptyExporter<>();
-  private static final Exporter<ForeignKey> NOOP_FK_EXPORTER = new EmptyExporter<>();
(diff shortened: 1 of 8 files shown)


src/main/java/com/google/cloud/spanner/hibernate/SpannerDialect.java
```diff
@@ -245,6 +250,53 @@ public String getForUpdateSkipLockedString(String aliases) {
         + " acquisition.");
   }
 
+  /* Unsupported Hibernate Exporters */
+
+  @Override
+  public Exporter<Sequence> getSequenceExporter() {
+    return new Exporter<Sequence>() {
```
ChengyuanZhao: Nice ! I love anonymous classes. 

Since these always just do nothing, would it be better to have  a private static instance ? rather than making new ones on each call ?

dzou: Good call, done.

ChengyuanZhao: if you don't give the type param like `Exporter NOOP_SEQUENCE_EXPORTER = new EmptyExporter<>();` you can actually the same instance for all 3 (though you do get a little implicitly-casting warning)

dzou: Done.

PR #36, opened 2019-03-06
Configure Dialect Testing to use Fork of hibernate-orm

dzou:
This configures the Dialect integration testing to use a local [fork of hibernate-orm](https://github.com/dzou/hibernate-orm), so we have the ability to make modifications to entities in the repo to get the tests to work.

### DIFF
--- a/testing/README.md
+++ b/testing/README.md
@@ -18,19 +18,12 @@ local maven in order to facilitate testing, so it will allow you to test local c
     Python 3.5.4rc1 
     ```
 
-2. Modify `databases.gradle` to include the correct JDBC URL to connect to your Spanner instance.
-You will need to specify the correct GCP project ID, Spanner instance, and table.
-
-    Example: 
-    
-    ```
-    'jdbc.url' : 'jdbc:cloudspanner://;Project=gcp_project_id;Instance=spanner_instance_name;Database=my_db'
-    ```
-
-3. Run `python3 run-tests.py`. Note that the script will run all the integration tests in the
+2. Run `python3 run-tests.py`. Note that the script will run all the integration tests in the
 directory. It is likely you may want to run a specific test for debugging; this can be done passing in a tests filter as a commandline argument to the script:
     
     Example:
     ```
     python3 run-tests.py SQLTest.test_sql_jpa_all_columns_scalar_query_example
     ```
+
+See [our fork of hibernate-orm](https://github.com/dzou/hibernate-orm) which contains the modified tests.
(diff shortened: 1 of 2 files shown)


dzou:
> What about just keeping the modified files, and layering them over the source repo? Or, are there too many of these changes that we would need to be maintained?

Yeah that would be tricky I think; I foresee many changes over many of the hibernate-orm entities so I don't think it will be feasible to do it that way. Like minimally it would be nice to have syntax highlighting when the files are connected in the project. So I think separate fork is preferable.

ChengyuanZhao:
> > What about just keeping the modified files, and layering them over the source repo? Or, are there too many of these changes that we would need to be maintained?
> 
> Yeah that would be tricky I think; I foresee many changes over many of the hibernate-orm entities so I don't think it will be feasible to do it that way. Like minimally it would be nice to have syntax highlighting when the files are connected in the project. So I think separate fork is preferable.

I agree. I think it would be a nightmare to read , write, or reason about layering on changed files only rather than working in a totally copied repo.

Also since hibernate 5 is totally done with no more releases we won't need to worry about the cloned one getting out of date.

meltsufin:
It sounds like the fork is preferable. The only thing that makes me a bit uneasy is that the fork is on a personal GitHub account, but I guess we can live with that for now. I wonder if importing the code into a branch in this repo is an option. :thinking:

dzou:
> It sounds like the fork is preferable. The only thing that makes me a bit uneasy is that the fork is on a personal GitHub account, but I guess we can live with that for now. I wonder if importing the code into a branch in this repo is an option.

Yeah if you want, we could just duplicate the hibernate-orm repo code under the testing/ directory in a separate branch. It could be easier; that approach would allow us to not have to clone the hibernate-orm fork from in the script.

dzou:
Hmm... actually I think we are actually best off with the current approach. I tried to copy/paste the hibernate-orm project into our project. However, because it is Gradle, Intellij was not able to compile the files in the subdirectory - i.e. it appears that it is not possible to open a project in intellij with uses both Maven and Gradle.

So I think the current approach with separate repositories is actually the better option.

meltsufin:
Maybe we should just switch to Gradle. :wink:

dzou:
> Maybe we should just switch to Gradle.

Yea I would be open to this idea. Does it make sense to do this in a later PR? Since I feel the changes will be fairly significant. (Created #39 to track)

meltsufin:
It should be possible to have both build systems in the same project in parallel. Why would the changes be so significant?

dzou:
> It should be possible to have both build systems in the same project in parallel. Why would the changes be so significant?

Fair enough; I assumed you meant to delete/replace the existing Maven setup as well. I'll try to add the parallel Gradle setup now.

dzou:
Going to close this in favor of #43.

Issue #17, opened 2019-01-25
TableExporter support for Collections and UNIQUE indexes 


ChengyuanZhao:
This may be pretty involved because in CREATE and DROP statements now any associated or interleaved indexes and child tables must also be dropped or created.

ChengyuanZhao:
NOTE: 

In Spanner the UNIQUE constraint does go in the CREATE TABLE statement. 

Instead you create an index on the column you want to be UNIQUE and that index is given the UNIQUE option, which then enforces uniqueness on the column. This is relevant for the OneToMany or ManyToOne relationship.

ChengyuanZhao:
Added collection support.

Unfortunately hibernate chooses the FK column names in the collection table and they are NOT the same as the names in the parent table as required by Spanner. So collection support is without interleaved support.

meltsufin:
Can you point to the code in Hibernate that does it? Is there any way to override it, since interleaved tables are in fact a perfect fit for collections.

ChengyuanZhao:
It looks like this is handled by `org.hibernate.cfg.annotations.CollectionBinder` where the behavior is hard-coded.

meltsufin:
Closing in favor of #140 and #165 .

PR #41, opened 2019-03-08
Support tests that use element collection

ChengyuanZhao:
related #17 
support element collections and helps w/ tests


ChengyuanZhao:
@meltsufin unfortunately there is a roadblock there. Hibernate chooses the column names for the parent-key columns in the collection table like `ParentEntity_ID1` rather than `ID1` as required by Spanner (Spanner requires the column names be the  same). Of course I could create the table with the original names, but later on Hibernate won't be aware of the different column names when querying.

meltsufin:
Sure, let's leave #17 to another PR.

PR #45, opened 2019-03-11
added tests for generated value sequence table

ChengyuanZhao:
added tests for generated ID table.

also changed mapping of numeric/decimal to string. However this requires JDBC support

### DIFF
--- a/src/main/java/com/google/cloud/spanner/hibernate/SpannerDialect.java
+++ b/src/main/java/com/google/cloud/spanner/hibernate/SpannerDialect.java
@@ -90,8 +90,8 @@ public class SpannerDialect extends Dialect {
     registerColumnType(Types.NCLOB, "STRING(MAX)");
     registerColumnType(Types.BLOB, "BYTES(MAX)");
 
-    registerColumnType(Types.DECIMAL, "STRING(MAX)");
-    registerColumnType(Types.NUMERIC, "STRING(MAX)");
+    registerColumnType(Types.DECIMAL, "FLOAT64");
+    registerColumnType(Types.NUMERIC, "FLOAT64");
 
     registerFunction("ANY_VALUE", new StandardSQLFunction("ANY_VALUE"));
     registerFunction("COUNTIF", new StandardSQLFunction("COUNTIF", StandardBasicTypes.LONG));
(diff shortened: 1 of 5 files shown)


src/main/java/com/google/cloud/spanner/hibernate/SpannerTableExporter.java
```diff
@@ -89,10 +89,18 @@ public SpannerTableExporter(SpannerDialect spannerDialect) {
             + " " + col.getSqlType(this.spannerDialect, metadata)
             + (col.isNullable() ? this.spannerDialect.getNullColumnString() : " not null")));
 
-    return new String[]{
-        MessageFormat.format(this.createTableTemplate, table.getQuotedName(),
+    ArrayList<String> statements = new ArrayList<>();
+    statements.add(MessageFormat.format(this.createTableTemplate, table.getQuotedName(),
             colsAndTypes.toString(),
-            primaryKeyColNames)};
+        primaryKeyColNames));
```
dzou: nit: indentation of `primaryKeyColNames` does not seem quite right.

src/main/java/com/google/cloud/spanner/hibernate/SpannerDialect.java
```diff
@@ -90,8 +90,8 @@ public SpannerDialect() {
     registerColumnType(Types.NCLOB, "STRING(MAX)");
     registerColumnType(Types.BLOB, "BYTES(MAX)");
 
-    registerColumnType(Types.DECIMAL, "FLOAT64");
-    registerColumnType(Types.NUMERIC, "FLOAT64");
+    registerColumnType(Types.DECIMAL, "STRING(MAX)");
```
dzou: Do you intend to include these changes with the sequence stuff?

src/main/java/com/google/cloud/spanner/hibernate/SpannerTableExporter.java
ChengyuanZhao: Fixed.

src/main/java/com/google/cloud/spanner/hibernate/SpannerDialect.java
ChengyuanZhao: reverted. I guess we should wait until we have a driver that supports these two types before we make changes

PR #46, opened 2019-03-11
Fix tests and initializes once

ChengyuanZhao:
Fix tests and initializes once

### DIFF
--- a/google-cloud-spanner-hibernate-testing/src/main/java/org/hibernate/userguide/model/Payment.java
+++ b/google-cloud-spanner-hibernate-testing/src/main/java/org/hibernate/userguide/model/Payment.java
@@ -6,7 +6,6 @@
  */
 package org.hibernate.userguide.model;
 
-import java.math.BigDecimal;
 import javax.persistence.Entity;
 import javax.persistence.GeneratedValue;
 import javax.persistence.Id;
@@ -26,7 +25,7 @@ public class Payment {
     @GeneratedValue
     private Long id;
 
-    private BigDecimal amount;
+    private Long amount;
 
     private boolean completed;
 
@@ -45,11 +44,11 @@ public class Payment {
         this.id = id;
     }
 
-    public BigDecimal getAmount() {
+    public Long getAmount() {
         return amount;
     }
 
-    public void setAmount(BigDecimal amount) {
+    public void setAmount(Long amount) {
         this.amount = amount;
     }
 
--- a/google-cloud-spanner-hibernate-testing/src/main/java/org/hibernate/userguide/model/Person.java
+++ b/google-cloud-spanner-hibernate-testing/src/main/java/org/hibernate/userguide/model/Person.java
@@ -77,12 +77,12 @@ import javax.persistence.Version;
         name = "find_person_by_name",
         query =
             "SELECT " +
-            "   p.id AS \"id\", " +
-            "   p.name AS \"name\", " +
-            "   p.nickName AS \"nickName\", " +
-            "   p.address AS \"address\", " +
-            "   p.createdOn AS \"createdOn\", " +
-            "   p.version AS \"version\" " +
+            "   p.id , " +
+            "   p.name , " +
+            "   p.nickName , " +
+            "   p.address , " +
+            "   p.createdOn , " +
+            "   p.version " +
             "FROM Person p " +
             "WHERE p.name LIKE :name",
         resultClass = Person.class
--- a/google-cloud-spanner-hibernate-testing/src/test/java/org/hibernate/jpa/test/BaseEntityManagerFunctionalTestCase.java
+++ b/google-cloud-spanner-hibernate-testing/src/test/java/org/hibernate/jpa/test/BaseEntityManagerFunctionalTestCase.java
@@ -28,7 +28,7 @@ import org.hibernate.jpa.HibernatePersistenceProvider;
 import org.hibernate.jpa.boot.spi.Bootstrap;
 import org.hibernate.jpa.boot.spi.PersistenceUnitDescriptor;
 import org.hibernate.testing.junit4.BaseUnitTestCase;
-import org.junit.After;
+import org.junit.AfterClass;
 import org.junit.Before;
 
 /**
@@ -45,10 +45,10 @@ public abstract class BaseEntityManagerFunctionalTestCase extends BaseUnitTestCa
 	private static final Dialect dialect = Dialect.getDialect();
 
 	private StandardServiceRegistryImpl serviceRegistry;
-	private SessionFactoryImplementor entityManagerFactory;
+	private static SessionFactoryImplementor entityManagerFactory;
 
-	private EntityManager em;
-	private ArrayList<EntityManager> isolatedEms = new ArrayList<EntityManager>();
+	private static EntityManager em;
+	private static ArrayList<EntityManager> isolatedEms = new ArrayList<EntityManager>();
 
 	protected Dialect getDialect() {
 		return dialect;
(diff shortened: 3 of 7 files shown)


google-cloud-spanner-hibernate-testing/src/test/java/org/hibernate/jpa/test/BaseEntityManagerFunctionalTestCase.java
```diff
@@ -286,16 +270,43 @@ private void releaseUnclosedEntityManager(EntityManager em) {
...
+	}
+
+	protected void doIfNotInitialized(Runnable runnable) {
+		if (!isInitialized) {
+			runnable.run();
+			isInitialized = true;
 		}
 	}
 
+	@Before
```
dzou: Does `@BeforeClass` not work here? Not too big of a fan of static variable null-check.

google-cloud-spanner-hibernate-testing/src/main/java/org/hibernate/userguide/model/Payment.java
```diff
@@ -26,7 +25,7 @@
     @GeneratedValue
     private Long id;
 
-    private BigDecimal amount;
+    private Long amount;
```
meltsufin: Have you looked into using the new JDBC Driver and if it supports `BigDecimal`? Also, maybe there is a way to just extend the Simba one to deal with `BigDecimal`. Otherwise, we need to document somewhere that `BigDecimal` is not supported. Let's start a page in this repo keeping track of unsupported features like this one.

google-cloud-spanner-hibernate-testing/src/main/java/org/hibernate/userguide/model/Person.java
```diff
@@ -77,12 +77,12 @@
         name = "find_person_by_name",
         query =
             "SELECT " +
-            "   p.id AS \"id\", " +
```
meltsufin: Another thing only we don't support?

google-cloud-spanner-hibernate-testing/src/test/java/org/hibernate/jpa/test/BaseEntityManagerFunctionalTestCase.java
```diff
@@ -62,20 +62,27 @@ protected StandardServiceRegistryImpl serviceRegistry() {
...
-		log.trace( "Building EntityManagerFactory" );
+	private static boolean isInitialized;
 
-		entityManagerFactory =  Bootstrap.getEntityManagerFactoryBuilder(
-				buildPersistenceUnitDescriptor(),
-				buildSettings()
-		).build().unwrap( SessionFactoryImplementor.class );
+	@AfterClass
+	@SuppressWarnings( {"UnusedDeclaration"})
+	public static void releaseResources() {
```
meltsufin: What are you trying to do here?

google-cloud-spanner-hibernate-testing/src/test/java/org/hibernate/jpa/test/BaseEntityManagerFunctionalTestCase.java
```diff
@@ -286,16 +270,43 @@ private void releaseUnclosedEntityManager(EntityManager em) {
...
+	protected void doIfNotInitialized(Runnable runnable) {
+		if (!isInitialized) {
+			runnable.run();
+			isInitialized = true;
 		}
 	}
 
+	@Before
+	@SuppressWarnings({"UnusedDeclaration"})
+	public void buildEntityManagerFactory() {
```
meltsufin: Can you explain a bit more what the changes are?

google-cloud-spanner-hibernate-testing/src/test/java/org/hibernate/userguide/sql/SQLTest.java
```diff
@@ -402,27 +409,6 @@ public void test_sql_hibernate_entity_associations_query_many_to_one_join_result
 		});
 	}
 
-	@Test
-	@RequiresDialect(H2Dialect.class)
-	@RequiresDialect(Oracle8iDialect.class)
-	@RequiresDialect(PostgreSQL82Dialect.class)
-	public void test_sql_jpa_entity_associations_query_one_to_many_join_example() {
```
meltsufin: Why would you need to delete this, if you it doesn't relate to the Spanner dialect anyway?

google-cloud-spanner-hibernate-testing/src/test/java/org/hibernate/userguide/sql/SQLTest.java
```diff
@@ -541,7 +502,7 @@ public void test_sql_hibernate_dto_query_example() {
 			Session session = entityManager.unwrap( Session.class );
 			//tag::sql-hibernate-dto-query-example[]
 			List<PersonSummaryDTO> dtos = session.createNativeQuery(
-				"SELECT p.id as \"id\", p.name as \"name\" " +
+				"SELECT p.id as id, p.name as name " +
```
meltsufin: Again, something we don't support but every other RDBMS does?

google-cloud-spanner-hibernate-testing/src/test/java/org/hibernate/userguide/sql/SQLTest.java
dzou: I think it's ok to delete this because this is just gonna be dead code as it is only run when those specific dialects are loaded.

google-cloud-spanner-hibernate-testing/src/test/java/org/hibernate/jpa/test/BaseEntityManagerFunctionalTestCase.java
ChengyuanZhao: This one here is the clean up so I changed it from After to AfterClass, but the AfterClass requires the method to be static. Before it would create and delete tables for each test, but since none of the tests actually change the data or tables we can do it all together at the end.

google-cloud-spanner-hibernate-testing/src/test/java/org/hibernate/jpa/test/BaseEntityManagerFunctionalTestCase.java
ChengyuanZhao: Hmm I tried making it BeforeClass but there are quite a lot of functions that are called along the line that all need to be made static as a result :( I think this would be preferable to that... What do you think? It ends up being like a dozen functions and members that need to be made to be static

google-cloud-spanner-hibernate-testing/src/main/java/org/hibernate/userguide/model/Person.java
ChengyuanZhao: Yes I see the usual databases support this for aliases that have spaces. 

Here is the issue for keeping a list of these sorts of things until we have a ref doc https://github.com/GoogleCloudPlatform/google-cloud-spanner-hibernate/issues/47

google-cloud-spanner-hibernate-testing/src/test/java/org/hibernate/userguide/sql/SQLTest.java
ChengyuanZhao: Yes this is the same quoted aliases that spanner doesn't support (but others do to support aliases with spaces, see #47)

google-cloud-spanner-hibernate-testing/src/test/java/org/hibernate/jpa/test/BaseEntityManagerFunctionalTestCase.java
dzou: Ah I see, that's fine then, you can just do it your current way.

google-cloud-spanner-hibernate-testing/src/main/java/org/hibernate/userguide/model/Payment.java
ChengyuanZhao: Interestingly the new driver claims to support all of the JDBC types except NUMERIC and DECIMAL (which corresponds to BigDecimal). (you can see this in the file `JdbcParameterStore.java`

dzou: If you want to try knut's driver, I built his shaded jar in some personal experiments here: https://github.com/dzou/knut-hibernate-experiment/tree/master/lib

google-cloud-spanner-hibernate-testing/src/test/java/org/hibernate/jpa/test/BaseEntityManagerFunctionalTestCase.java
ChengyuanZhao: Added explanation in code, PTAL

google-cloud-spanner-hibernate-testing/src/main/java/org/hibernate/userguide/model/Person.java
ChengyuanZhao: @meltsufin @dzou any outstanding change requests for this test file?

dzou: Nope, I am good.

Issue #52, opened 2019-03-18
Bring over hibernate ORM integration tests


ChengyuanZhao:
- [x] https://github.com/hibernate/hibernate-orm/blob/master/documentation/src/test/java/org/hibernate/userguide/sql/SQLTest.java
- [x] https://github.com/hibernate/hibernate-orm/blob/master/documentation/src/test/java/org/hibernate/userguide/hql/HQLTest.java
- [x] https://github.com/hibernate/hibernate-orm/tree/master/documentation/src/test/java/org/hibernate/userguide/inheritance
- [x] https://github.com/hibernate/hibernate-orm/tree/master/documentation/src/test/java/org/hibernate/userguide/fetching
- [x] https://github.com/hibernate/hibernate-orm/tree/master/documentation/src/test/java/org/hibernate/userguide/collections
- [x] https://github.com/hibernate/hibernate-orm/tree/master/documentation/src/test/java/org/hibernate/userguide/persister
- [x] https://github.com/hibernate/hibernate-orm/tree/master/documentation/src/test/java/org/hibernate/userguide/schema
- [x] https://github.com/hibernate/hibernate-orm/tree/master/documentation/src/test/java/org/hibernate/userguide/transactions
- [x] https://github.com/hibernate/hibernate-orm/tree/master/documentation/src/test/java/org/hibernate/userguide/mapping

These are the topics that we could bring over. pick and choose ones to work on @dmitry-s @dzou ?

dzou:
Sure thing, I'll take the HQL Tests

ChengyuanZhao:
I'll take the `collections` test directory

ChengyuanZhao:
@dzou @meltsufin 

Many of the collection-directory tests modify the state of the tables and need to have those tables deleted and recreated and sometimes as tables with the same name but different actual POJO classes...

For those classes that have these issues we will need to go back to what hibernate does (recreate everything after each test)

dzou:
I see. It might be OK since I see that the Collection tests appear to only use 2 entities (like Person + Phone). This means each test would only need to recreate/delete 2 tables which is not too bad. So we can just try to add these tests to the CI and see what effects it has on the run time; if it's not too bad then we can keep.

Maybe if our CI grows to take too long (like taking +12 hours) we can consider something else such as splitting certain sets of tests to be run manual-only.

ChengyuanZhao:
I'll take the `fetching` tests now

[...]

Issue #47, opened 2019-03-12
Unsupported Features List


ChengyuanZhao:
This is a list of features we don't support that I will keep here until we have a place in documentation to put them:

- Simba driver AND new driver don't support jdbc NUMERIC or DECIMAL. This means types like BigDecimal cannot be supported regardless of how we want to store them in Spanner because the driver doesn't even attempt to
- several integration tests use aliases in quotes. This is not supported in Spanner, but is supported in traditional databases (for aliases that have spaces in them)

dzou:
- Spanner does not support the [WHERE ... ANY or WHERE ... ALL clauses](https://www.w3schools.com/sql/sql_any_all.asp).

meltsufin:
@dzou  Have you tried it? Interestingly, `ANY` and `ALL` are [reserved](https://cloud.google.com/spanner/docs/lexical) keywords in Spanner.

dzou:
> @dzou Have you tried it? Interestingly, `ANY` and `ALL` are [reserved](https://cloud.google.com/spanner/docs/lexical) keywords in Spanner.

Yeah, pretty sure it doesn't work. When I try something like:

```
SELECT *
FROM Employee
WHERE name = ANY (SELECT name from Employee)
```

I get error "unexpected keyword ANY".

I suspect they reserved the ANY/ALL keywords because they might expand their features to include it in the future.

ChengyuanZhao:
closing since we're apparently using separate issues for each

PR #53, opened 2019-03-19
Enable HQL Tests

dzou:
This enables the HQL tests and makes a handful of corrections to the Dialect.

Contributes to #52.


google-cloud-spanner-hibernate-testing/src/test/java/org/hibernate/userguide/hql/HQLTest.java
```diff
@@ -0,0 +1,2293 @@
...
+        "from Person p " +
+        "where p.name like 'Joe'", Person.class)
+      .getResultList();
+      //end::hql-string-literals-example[]
+    });
+  }
+
+  @Test
+  @Ignore
+  // TODO: Test fails due to quote literals in string: 'Joe''s'; figure out if there
```
meltsufin: It's not clear in this file which parts are original, and which ones you changed. Maybe you can split your changes into a separate commit?

Also, this TODO should probably be an issue we're tracking.

Finally, did you have to delete any tests like the ones involving `= ANY (SELECT ...)`?

dzou: I'll make a separate PR with the test annotated with `@Ignore` to visualize changes

google-cloud-spanner-hibernate-testing/src/test/java/org/hibernate/userguide/hql/HQLTest.java
```diff
@@ -1469,24 +1432,6 @@ public void test_hql_collection_expressions_example_4() {
...
-	public void test_hql_collection_expressions_example_5() {
-		doInJPA( this::entityManagerFactory, entityManager -> {
-			Call call = entityManager.createQuery( "select c from Call c", Call.class).getResultList().get( 0 );
-			Phone phone = call.getPhone();
-			//tag::hql-collection-expressions-example[]
-
-			List<Person> persons = entityManager.createQuery(
-				"select p " +
-				"from Person p " +
-				"where :phone = some elements ( p.phones )", Person.class )
```
ChengyuanZhao: this  `= som elements ( ....` becomes ` = ANY (... ` ?

google-cloud-spanner-hibernate-testing/src/test/java/org/hibernate/userguide/hql/HQLTest.java
```diff
@@ -1510,28 +1455,13 @@ public void test_hql_collection_expressions_example_7() {
...
 
-	@Test
-	public void test_hql_collection_expressions_example_8() {
-		doInJPA( this::entityManagerFactory, entityManager -> {
-			//tag::hql-collection-expressions-example[]
-
-			List<Phone> phones = entityManager.createQuery(
-				"select p " +
-				"from Phone p " +
-				"where current_date() > all elements( p.repairTimestamps )", Phone.class )
```
ChengyuanZhao: I assume this becomes the ` = ALL ( .. ` ?

google-cloud-spanner-hibernate-testing/src/test/java/org/hibernate/userguide/hql/HQLTest.java
```diff
@@ -2041,22 +1947,6 @@ public void test_hql_like_predicate_example_2() {
...
-	@Test
-	public void test_hql_like_predicate_escape_example() {
-
-		doInJPA( this::entityManagerFactory, entityManager -> {
-			//tag::hql-like-predicate-escape-example[]
-			// find any person with a name starting with "Dr_"
-			List<Person> persons = entityManager.createQuery(
-				"select p " +
-				"from Person p " +
-				"where p.name like 'Dr|_%' escape '|'", Person.class )
```
ChengyuanZhao: this tests `LIKE` ?  That is in Cloud Spanner though

google-cloud-spanner-hibernate-testing/src/test/java/org/hibernate/userguide/hql/HQLTest.java
```diff
@@ -2073,25 +1963,6 @@ public void test_hql_between_predicate_example_1() {
...
-	@RequiresDialect(MySQL5Dialect.class)
-	public void test_hql_between_predicate_example_2() {
-
-		doInJPA( this::entityManagerFactory, entityManager -> {
-			//tag::hql-between-predicate-example[]
-
-			List<Person> persons = entityManager.createQuery(
-				"select p " +
-				"from Person p " +
-				"where p.createdOn between '1999-01-01' and '2001-01-02'", Person.class )
```
ChengyuanZhao: What does this `between` translate to ?

google-cloud-spanner-hibernate-testing/src/test/java/org/hibernate/userguide/hql/HQLTest.java
dzou: 1. Done.
2. Done. Made this issue to track: https://github.com/GoogleCloudPlatform/google-cloud-spanner-hibernate/issues/55
3. Yes, tests using ANY, ALL, and SOME were deleted.

google-cloud-spanner-hibernate-testing/src/test/java/org/hibernate/userguide/hql/HQLTest.java
```diff
@@ -2330,56 +2200,6 @@ public void test_hql_group_by_example_2() {
 		});
 	}
 
-	@Test
-	@RequiresDialect(H2Dialect.class)
-	@RequiresDialect(PostgreSQL81Dialect.class)
-	@RequiresDialect(MySQL5Dialect.class)
-	public void test_hql_group_by_example_3() {
```
ChengyuanZhao: Would this one also work in Spanner? I know this one specifies it only works for some dialects, but it mentions both postgres and mysql, which feel like pretty general SQL to me

google-cloud-spanner-hibernate-testing/src/test/java/org/hibernate/userguide/hql/HQLTest.java
```diff
@@ -1469,24 +1432,6 @@ public void test_hql_collection_expressions_example_4() {
 		});
 	}
 
-	@Test
```
meltsufin: Instead of deleting tests that don't work, I would just put `@Ignore` on them with a short description of why.

dzou: Done.

google-cloud-spanner-hibernate-testing/src/test/java/org/hibernate/userguide/hql/HQLTest.java
dzou: No, turns out `SOME` is [its own keyword](https://www.geeksforgeeks.org/sql-some/). Spanner does not yet support "SOME".

google-cloud-spanner-hibernate-testing/src/test/java/org/hibernate/userguide/hql/HQLTest.java
dzou: Yup, correct.

google-cloud-spanner-hibernate-testing/src/test/java/org/hibernate/userguide/hql/HQLTest.java
dzou: This one doesn't work because it uses `like` with `escape` to specify a custom character to use as an escape delimiter. The escape half is not yet supported in Spanner.

google-cloud-spanner-hibernate-testing/src/test/java/org/hibernate/userguide/hql/HQLTest.java
dzou: Seems like it's just the the vanilla Between keyword: https://www.w3schools.com/sql/sql_between.asp

It works in Spanner.

google-cloud-spanner-hibernate-testing/src/test/java/org/hibernate/userguide/hql/HQLTest.java
dzou: Hmm it didn't work; seemed to fail at the group by. This one looks like it might be able to make work with some adjustments to Dialect; created issue to track: https://github.com/GoogleCloudPlatform/google-cloud-spanner-hibernate/issues/56

meltsufin:
@dzou Are we adding all of these unsupported features because of which you're ignoring tests to the tracking issue(s)?
Also, it might be useful for the future to reference the issue(s) in the comments, so that when the feature is supported by Spanner, we can easily find the test to re-enable.

dzou:
> @dzou Are we adding all of these unsupported features because of which you're ignoring tests to the tracking issue(s)?
> Also, it might be useful for the future to reference the issue(s) in the comments, so that when the feature is supported by Spanner, we can easily find the test to re-enable.

Yes, the unsupported features have been added to the tracking issue #47.

Hmm, since there is only one issue #47 tracking unsupported features, we would be adding "#47" everywhere there is an Ignore, so I don't think it would be too useful to add; would just be redundant. There are 2 tests which I have identified that could potentially be fixed, and these have separate issues created for them.

meltsufin:
I think it would be easier to track if we had separate issues rather than lumping everything into #47. We can use a tag to group them.

dzou:
> I think it would be easier to track if we had separate issues rather than lumping everything into #47. We can use a tag to group them.

Done. PTAL.

meltsufin:
Perfect! Thanks @dzou!

PR #74, opened 2019-04-02
Add Mapping tests to repo with no changes

dzou:
This adds the Mapping tests to our repo with no changes.


dzou:
> These are going to be fun! :)

Very fun indeed.... :joy:

PR #78, opened 2019-04-03
Fix the Mapping Tests

dzou:
This applies fixes necessary to make `mapping` tests pass.


google-cloud-spanner-hibernate-testing/src/test/java/org/hibernate/userguide/mapping/basic/ParentTest.java
```diff
@@ -45,7 +46,7 @@ public void testLifecycle() {
 		//tag::mapping-Parent-fetching-example[]
 		doInJPA( this::entityManagerFactory, entityManager -> {
 
-			City cluj = entityManager.find( City.class, 1L );
+			City cluj = entityManager.find( City.class, 666L );
```
ChengyuanZhao: nice ID

google-cloud-spanner-hibernate-testing/src/test/java/org/hibernate/userguide/mapping/basic/ParentTest.java
```diff
@@ -97,11 +98,10 @@ public void setCity(City city) {
 
 	//tag::mapping-Parent-example[]
 
-	@Entity(name = "City")
+	@Entity(name = "City_BasicParentTest")
 	public static class City {
 
 		@Id
-		@GeneratedValue
```
ChengyuanZhao: does the generated value cause problems? Do some tests rely on a static id (the sequence table being reset each test) ?

For those cases I changed the queries instead of removing the generated value, but either way works .

dzou: Right, there were no problems using GeneratedValue; however, the test was doing a `entitymanager.getById` with `id == 1` and this assumes the sequence table being reset with each test. So I just removed generatedValue so that we can set the ID manually; figure this would be easier.

ChengyuanZhao: Yea I came across a few of those situations, but I replaced `1` with `(select min(id) from hibernate_sequence)`. Anyways LGTM

google-cloud-spanner-hibernate-testing/src/test/java/org/hibernate/userguide/mapping/basic/ParentTest.java
dzou: :open_mouth:

PR #84, opened 2019-04-04
Fix Test Configuration

dzou:
This PR makes the following changes to fix the build:

1. Enable running tests only during merge.
2. Import BaseNonConfigCoreFunctionalTestCase to repo to modify the hibernate mode to "UPDATE". this is the last parent test class which was still creating and dropping tables.
3. Suppress log output to resolve error "The job exceeded the maximum log length, and has been terminated." https://api.travis-ci.org/v3/job/515405979/log.txt


google-cloud-spanner-hibernate-testing/src/test/java/org/hibernate/userguide/mapping/embeddable/EmbeddableImplicitOverrideTest.java
```diff
@@ -9,7 +9,7 @@
 import org.hibernate.annotations.NaturalId;
 import org.hibernate.boot.MetadataBuilder;
 import org.hibernate.boot.model.naming.ImplicitNamingStrategyComponentPathImpl;
-import org.hibernate.testing.junit4.BaseNonConfigCoreFunctionalTestCase;
+import org.hibernate.jpa.test.BaseNonConfigCoreFunctionalTestCase;
```
meltsufin: What's the difference in this one vs the one that was there before? Same name.

google-cloud-spanner-hibernate-testing/src/test/java/org/hibernate/userguide/collections/BasicTypeCollectionTest.java
```diff
@@ -58,13 +58,6 @@ protected Configuration constructAndConfigureConfiguration() {
 		return configuration;
 	}
 
-	@Before
-	public void cleanTables() {
```
meltsufin: Why is this no longer necessary?

google-cloud-spanner-hibernate-testing/src/test/java/org/hibernate/userguide/mapping/embeddable/EmbeddableImplicitOverrideTest.java
dzou: This is a parent test class in the Hibernate libs; we copied it over to our repo under that new package name "org.hibernate.jpa.test" and modified the `HBM2DDL_AUTO` setting in it from 'create-drop' to update. This is what allows us to do the thing where we don't recreate tables.

google-cloud-spanner-hibernate-testing/src/test/java/org/hibernate/userguide/collections/BasicTypeCollectionTest.java
dzou: This class now inherits from the modified `BaseCoreFunctionalTestCase` that we copied over which contains the cleanTables logic, so this is now redundant. The children classes no longer need clean table logic individually.

.travis.yml
```diff
@@ -17,12 +17,12 @@ install:
...
 
 before_install:
-  - if [ "$TRAVIS_SECURE_ENV_VARS" == "true" ]; then
+  - if [ "$TRAVIS_SECURE_ENV_VARS" == "true" ] && [ "$TRAVIS_PULL_REQUEST" == "false" ]; then
       openssl aes-256-cbc -K $encrypted_5bb5dbfd81fd_key -iv $encrypted_5bb5dbfd81fd_iv -in travis.tar.gz.enc -out travis.tar.gz -d;
       tar -xzf travis.tar.gz;
       export GOOGLE_APPLICATION_CREDENTIALS=$TRAVIS_BUILD_DIR/travis/admin.json;
       export GOOGLE_CLOUD_PROJECT=cloud-spanner-hibernate-ci;
-      export INTEGRATION_TEST_FLAGS="-Dit.hibernate=true";
+      export INTEGRATION_TEST_FLAGS="-Dit.hibernate=true -Dhibernate.show_sql=false";
```
ChengyuanZhao: If a test class doesn't print anything for 10 minutes (like the very-long HQLTest class) then it will get killed!

dzou:
> Where are you suppressing log output?

Log output is suppressed in travis.yml via command line property in the integration tests flags.

google-cloud-spanner-hibernate-testing/src/test/java/org/hibernate/jpa/test/BaseCoreFunctionalTestCase.java
```diff
@@ -132,6 +134,14 @@ private String getDeleteQuery(String tableName) {
 
   // before/after test class ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+  @BeforeClass
```
ChengyuanZhao: ooh I think we forgot to call cleanTables() on line 382 of this file :S

.travis.yml
meltsufin: Is most of the logging really coming from the actual SQL statements being printed out?

dzou: Yeah so in the log file I linked, you can see all the statements printed. All of those SQL statements are suppressed with the flag. https://api.travis-ci.org/v3/job/515405979/log.txt

You can see at the end of the log file, our build failed because "The job exceeded the maximum log length, and has been terminated." This PR attempts to solve the problem by suppressing log output; however, maybe this is suppressing too much output...

So far we've identified the following ways our build could fail:
- If no log output is produced in 10 minutes, the build fails
- If the build takes more than 50 minutes, it fails
- If too much log output is produced, the build fails

meltsufin: I see a lot of log output from the DEBUG level. Can we just change the log level to INFO?
I think keeping the sql strings in the logs is useful though, unless it is the primary cause of the log getting too large, and not the DEBUG level.

PR #91, opened 2019-04-10
test fixes

ChengyuanZhao:
Also fixed ref doc. those other bulk id strategies produce incompatible syntax in queries.


ChengyuanZhao:
@dzou PTAL

PR #92, opened 2019-04-10
Initial Port of our Dialect to Hibernate 6

dzou:
This is the initial port of our dialect + table exporter to Hibernate 6.

I have created a branch called hibernate6 where we can iterate on changes.
**Note: this is not the permanent location of the dialect; this is just put here to leverage our test infrastructure and have a place for development.**

Makes fixes to unit tests as well.


google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/SpannerDialectTableExporter.java
```diff
@@ -49,84 +49,78 @@
...
 
     if (table.hasPrimaryKey()) {
       // a typical table that corresponds to an entity type
       keyColumns = table.getPrimaryKey().getColumns();
-    } else if (containingTable != null) {
-      // a table that is actually an element collection property
-      keyColumns = table::getColumnIterator;
+    } else if (table.getForeignKeys().size() > 0) {
+      // a table with no PK's but has FK's; often corresponds to element collection properties
+      keyColumns = table.getPhysicalColumns();
```
dzou: In hibernate 6, the table exporter does not get access to the metadata object as before, so one cannot do similar operation as `getContainingTableForCollection(...)`.

As a substitute, I modified it such that if a table does not have a PK but has at least 1 FK, then all of the columns in that table will be used as a PK for that table in Spanner. Not sure if we want to take this approach or go another way.

google-cloud-spanner-hibernate-dialect/src/test/java/com/google/cloud/spanner/hibernate/GeneratedSelectStatementsTests.java
```diff
@@ -80,13 +81,13 @@ public void setup() {
...
-    testStatementTranslation(x -> {
+    List<String> statements = getStatementTranslation(x -> {
       Query q = x.createQuery("select s from SubTestEntity s")
           .setFirstResult(8).setMaxResults(15);
       q.setLockMode(LockModeType.PESSIMISTIC_READ);
       q.list();
-    }, "select subtestent0_.id as id1_1_, subtestent0_.id1 as id2_1_, "
-        + "subtestent0_.id2 as id3_1_ from SubTestEntity subtestent0_ limit ? offset ?");
+    });
+    assertThat(statements.get(0)).doesNotContain("for update");
```
ChengyuanZhao: Why remove the actual string? having the actual expected string tests for several other things other than just not having "for update" such as  the  offset and limit appearing

dzou: Yeah I wanted to keep it, but unfortunately Hibernate 6 behaves non-deterministically; i.e. the assertion failed sometimes because half the time it would rearrange the terms in the query in one way, half the time in another.

I changed it here since we don't really care too much about the actual contents of the query but the existence/non-existence certain keywords.

google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/SpannerDialect.java
```diff
@@ -95,170 +95,398 @@ public SpannerDialect() {
...
+        .setInvariantType(StandardSpiBasicTypes.BOOLEAN)
+        .setExactArgumentCount(1)
+        .register();
+    registry.namedTemplateBuilder("STRING_AGG")
+        .setInvariantType(StandardSpiBasicTypes.STRING)
+        .setArgumentCountBetween(1, 2)
+        .register();
+
+    // Mathematical Functions
+    CommonFunctionFactory.abs(registry);
```
meltsufin: I don't see this one in the list above that you replaced from ORM 5.
It looks like you referenced the list of functions provided [here](https://cloud.google.com/spanner/docs/functions-and-operators). ORM 5 list doesn't seem to follow the same approach. Can we ensure consistency?

google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/SpannerDialect.java
```diff
@@ -598,16 +818,34 @@ public void lock(Serializable id, Object version, Object object, int timeout,
   static class EmptyExporter<T extends Exportable> implements Exporter<T> {
 
     @Override
-    public String[] getSqlCreateStrings(T exportable, Metadata metadata) {
+    public String[] getSqlCreateStrings(
+        T exportable, JdbcServices jdbcServices) {
```
meltsufin: nit: Why arguments on a new line?
Seems like unnecessary diff fuzz. Also, a bunch more instances below.

google-cloud-spanner-hibernate-dialect/src/test/java/com/google/cloud/spanner/hibernate/GeneratedSelectStatementsTests.java
meltsufin: This may become a much bigger problem down the road for testing. Can we find out why this is happening, and whether there is a way to disable this non-determinism?

google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/SpannerDialect.java
meltsufin: @dmitry-s How did you approach the original list in ORM 5 dialect to ensure completeness?

google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/SpannerDialect.java
dzou: Yeah this was done by the intellij auto formatting. I'll look into how to import Hibernate checkstyle for us; that would be ideal moving forward.

google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/SpannerDialect.java
dmitry-s: I got the list of functions from the documentation.

dmitry-s: @meltsufin You can't see it here because it is inherited from the parent class.

meltsufin: @dzou Did the parent class change for ORM 6?

dzou: Same parent class, but they have different ways of implementing function mappings.

Hibernate 5 way: https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java#L162

Hibernate 6 way: https://github.com/hibernate/hibernate-orm/blob/wip/6.0/hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java#L283

Pretty sure both dialects are equivalent. I did this port by writing a script and running a regex expression on the Spanner Hibernate 5 mappings and then converted them to Hibernate 6 new style. Then manually verified + compared with [postgres Hibernate 6 dialect](https://github.com/hibernate/hibernate-orm/blob/wip/6.0/hibernate-core/src/main/java/org/hibernate/dialect/PostgreSQL81Dialect.java#L113). That's where the CommonFunctionRegistry stuff was used.

dzou:
> I think we need to introduce checkstyle that is compatible with Hibernate's style. Otherwise, fixing all this code is going to get really painful in the future when a lot more accumulates.

Done. I imported the hibernate 6 checkstyle into this branch. So creating a PR with our dialect for Hibernate 6 will be much easier.

google-cloud-spanner-hibernate-dialect/src/test/java/com/google/cloud/spanner/hibernate/GeneratedSelectStatementsTests.java
dzou: The cause appears to be related to the new SQM strategy they implemented for Hibernate 6. I couldn't find a way to disable this non-determinism; the source is likely something fundamental in their algorithm.

The non-determinism is not going to be a significant problem for testing - as long as the query generated is valid it's okay.

The only type of test this precludes us from doing are unit tests in which we input some HQL query and expect some SQL to come out - but moving forward I don't see us adding much more of these kind of tests since we are past the development stage, and we really care about verifying the queries against Real Spanner rather than just comparing the actual sql string against some expected one.

pom.xml
```diff
@@ -98,16 +98,6 @@
           </dependency>
         </dependencies>
         <executions>
-          <execution>
```
meltsufin: Should we replace it with an execution with the hibernate style?

google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/SpannerDialect.java
meltsufin: cool!

pom.xml
dzou: Yup, this was actually already being done but I renamed the files for clarity. I put all of the Hibernate checks into custom-checks.xml. Now just renamed the xml files to be more clear.

PR #95, opened 2019-04-24
Repairs the Travis CI to run all tests

dzou:
This repairs the Travis CI to run all tests.

- Driver is provided in the travis.tar.gz.enc
- The bulk ID strategy is set to "no-op" because the delete statements it generates are redundant to the statements produced by the @After cleanup of the BaseEntityManagerFunctionalTestCase; this should resolve timeout issues.


.travis.yml
```diff
@@ -18,11 +18,12 @@ install:
 
 before_install:
   - if [ "$TRAVIS_SECURE_ENV_VARS" == "true" ]; then
-      openssl aes-256-cbc -K $encrypted_5bb5dbfd81fd_key -iv $encrypted_5bb5dbfd81fd_iv -in travis.tar.gz.enc -out travis.tar.gz -d;
+      openssl aes-256-cbc -K $encrypted_d2c20fa49002_key -iv $encrypted_d2c20fa49002_iv -in travis.tar.gz.enc -out travis.tar.gz -d;
```
meltsufin: Why new key?

meltsufin:
The build seems still broken though

meltsufin:
@dzou Any way to disable the debug level logging?

dzou:
> @dzou Any way to disable the debug level logging?

Hmm I don't think suppressing the debug level logging will solve the problem; you can see from past runs that the large majority (90%+) of the log is contributed by the SQL which is why I have to suppress the sql that gets printed out: https://api.travis-ci.org/v3/job/520484879/log.txt

Debugging the tests should still be easy though because it is still easy to see which tests fails; and then running them locally can reveal the source of the error.

.travis.yml
dzou: Hmm when I ran `travis encrypt-file` it gave me a new command to copy-paste. Maybe they create a new key for each file you encrypt?

dzou:
All tests our now passing - merging in.

PR #96, opened 2019-04-25
Separated giant HQL Tests into more classes so each runs for less

ChengyuanZhao:
I separated the giant HQL test file into more files so each runs for less than 10 minutes.
disabled the show_sql so it doesn't crash the travis CI


ChengyuanZhao:
@dzou i saw you actually had a much better idea in the other PR of printing test name after each test!

dzou:
> @dzou i saw you actually had a much better idea in the other PR of printing test name after each test!

Yeah haha. The CI should be fixed now, so these changes may not be necessary!

PR #98, opened 2019-04-25
test fixes

ChengyuanZhao:
test fix 

I think the noop bulk id handler was doing too little in some cases and while the first run of tests would succeed the following runs wouldn't have cleaned up the table properly.

also i reverted a table rename, which was causing failures on second-runs because i think there were some issues or tests leaving it in some unexpected state.


dzou:
Hmm, might need to keep the noopbulkidhandler and then augment the `cleanupTables` to do the job more thoroughly if it is taking more than 50 minutes.

I changed it to the noopbulkidhandler b/c I saw it was generating a lot of delete statements that cleared tables row-by-row which was taking a lot of time; I thought the bulkIdhandler was redundant to the `cleanupTables` method we have which was doing `delete from X where 1=1`.

ChengyuanZhao:
@dzou  I think hibernate is pretty convoluted here. I was seeing lots of issues with "ROW ALREADY EXISTS" implying that things weren't being cleaned up, but these issues went away when I went back to the old method. 

Let's see if this runs quickly enough. I think this non-parallel row-by-row delete was too slow , but the slightly modified version we had for a while ran in like 32 minutes. I'll wait and see and let you know.

dzou:
> @dzou I think hibernate is pretty convoluted here. I was seeing lots of issues with "ROW ALREADY EXISTS" implying that things weren't being cleaned up, but these issues went away when I went back to the old method.
> 
> Let's see if this runs quickly enough. I think this non-parallel row-by-row delete was too slow , but the slightly modified version we had for a while ran in like 32 minutes. I'll wait and see and let you know.

Sounds good :+1:

PR #99, opened 2019-04-25
Test fixes

ChengyuanZhao:
test fixes

Mostly fixes are hunting down final issues with cleanup between tests. 
The fixes:

- on some tests we observed repeated ALREADY EXISTS errors even though the program output was showing that DELETE FROM commands were being sent that should've cleaned out the tables. It turned out in those cases we had sent delete commands to clean out not-real tables (but appear as tables in the hibernate metamodel). We were catching these errors but the simple act of attempting it somehow made later delete queries ignored. So the solution there was to issue these clean-table queries individually as their own transactions instead of all together. 

- The noop bulk Id strat seemed to not submit delete operations in cases when some where needed. perhaps our understanding of exactly where hibernate calls the bulk id strat was incomplete (it is frankly very convoluted to figure out ). So we reverted to using essentially the hibernate-provided bulk id strat that worked. However by looking at the program output it looked like in a long series of tests Hibernate emitted repeated delete calls that were identical! So I added a little check to not submit repeated identical deletes. It's not clear why this was happening (this would require a deep examination of hibernate I think) 

- I changed the format_sql setting to false. This causes queries to be printed all on a single line instead of broken over many lines. I find this actually much easier to read because the "formatted" way often had single words on each line ..


ChengyuanZhao:
@dzou  PTAL, it took 17 minutes or so and finished successfully ! let's see if it'll stick for the master branch ?

ChengyuanZhao:
@meltsufin  ah my bad! I had to add that file back for local testing but forgot to exclude it from the commit . also updated description

PR #134, opened 2019-10-22
Remake Travis creds for Spanner Integration tests

dzou:
This remakes the Travis secret in order to reenable the integration tests.

I think the root issue was that our creds were available on travis-ci.org but not travis-ci.com. This remakes the creds for travis-ci.com which should make the creds available there again and reenable the tests.


dzou:
Gonna submit this; looks like all the tests pass, however there appears to be an issue with Sonar configuration. Will followup on this.

PR #144, opened 2019-11-05
Add code to benchmark basic Hibernate operations

dzou:
This adds code to begin benchmarking some basic Hibernate operations.

More operations will be added, such as bulk DDL; this is not a complete list.

### DIFF
--- a/google-cloud-spanner-hibernate-performance-testing/src/main/java/com/google/cloud/spanner/hibernate/ClientLibraryOperations.java
+++ b/google-cloud-spanner-hibernate-performance-testing/src/main/java/com/google/cloud/spanner/hibernate/ClientLibraryOperations.java
@@ -23,9 +23,11 @@ import com.google.cloud.Date;
 import com.google.cloud.spanner.DatabaseAdminClient;
 import com.google.cloud.spanner.DatabaseClient;
 import com.google.cloud.spanner.DatabaseId;
+import com.google.cloud.spanner.ErrorCode;
 import com.google.cloud.spanner.Mutation;
 import com.google.cloud.spanner.ResultSet;
 import com.google.cloud.spanner.Spanner;
+import com.google.cloud.spanner.SpannerException;
 import com.google.cloud.spanner.SpannerOptions;
 import com.google.cloud.spanner.Statement;
 import com.google.spanner.admin.database.v1.UpdateDatabaseDdlMetadata;
@@ -37,11 +39,15 @@ import java.util.Collections;
 import java.util.List;
 import java.util.UUID;
 import java.util.stream.Collectors;
+import org.apache.log4j.Logger;
 
 /**
  * Runs common operations (reads, writes) on Spanner using the Spanner Client Libraries.
  */
 public class ClientLibraryOperations {
+
+  private static final Logger LOGGER = Logger.getLogger(ClientLibraryOperations.class);
+
   private static final String DDL_SMALL = "/ddl_strings/airport_ddl.txt";
   private static final String DDL_LARGE = "/ddl_strings/bulk_ddl.txt";
 
@@ -75,7 +81,16 @@ public class ClientLibraryOperations {
    * Resets the test Spanner database.
    */
   public void resetTestDatabase() {
-    databaseAdminClient.dropDatabase(INSTANCE_NAME, DATABASE_NAME);
+    try {
+      databaseAdminClient.dropDatabase(INSTANCE_NAME, DATABASE_NAME);
+    } catch (SpannerException e) {
+      if (e.getErrorCode() != ErrorCode.NOT_FOUND) {
+        throw new RuntimeException(e);
+      } else {
+        LOGGER.info("Skipping dropping the test database because it was not found.");
+      }
+    }
+
     try {
       databaseAdminClient.createDatabase(
           INSTANCE_NAME, DATABASE_NAME, Collections.EMPTY_LIST).get();
--- a/google-cloud-spanner-hibernate-performance-testing/src/main/java/com/google/cloud/spanner/hibernate/ClientLibraryOperations.java
+++ b/google-cloud-spanner-hibernate-performance-testing/src/main/java/com/google/cloud/spanner/hibernate/ClientLibraryOperations.java
@@ -72,22 +72,16 @@ public class ClientLibraryOperations {
   }
 
   /**
-   * Creates a database.
+   * Resets the test Spanner database.
    */
-  public void createTestDatabase() {
+  public void resetTestDatabase() {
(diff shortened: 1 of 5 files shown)


google-cloud-spanner-hibernate-performance-testing/src/main/java/com/google/cloud/spanner/hibernate/ClientLibraryOperations.java
```diff
@@ -72,9 +72,10 @@ public ClientLibraryOperations() {
   }
 
   /**
-   * Creates a database.
+   * Resets the test Spanner database.
    */
-  public void createTestDatabase() {
+  public void resetTestDatabase() {
+    databaseAdminClient.dropDatabase(INSTANCE_NAME, DATABASE_NAME);
```
elefeint: My recollection is that dropping a nonexistent database results in exception log clutter. If it still does that, might want to check if database exists first.

google-cloud-spanner-hibernate-performance-testing/src/main/java/com/google/cloud/spanner/hibernate/HibernateDriver.java
```diff
@@ -0,0 +1,60 @@
...
+        () -> hibernateOperations.insertRows(1000),
+        "Insert 1000 rows by saving Hibernate entities.");
+
+    BenchmarkUtil.benchmark(
+        () -> hibernateOperations.updateRows(1),
+        "Update 1 row by saving Hibernate entities.");
+
+    BenchmarkUtil.benchmark(
+        () -> hibernateOperations.updateRows(1000),
+        "Update 1000 row by saving Hibernate entities.");
```
elefeint: s/row/rows

google-cloud-spanner-hibernate-performance-testing/src/main/java/com/google/cloud/spanner/hibernate/HibernateOperations.java
```diff
@@ -0,0 +1,94 @@
...
+  void insertRows(int rowCount) {
+    Session session = this.sessionFactory.openSession();
+    session.setFlushMode(FlushModeType.COMMIT);
+
+    Transaction tx = session.beginTransaction();
+
+    for (int i = 0; i < rowCount; i++) {
+      Airport airport = new Airport(
+          "The Airport",
+          i + " main street",
```
elefeint: Hibernate test setup incurs a slight concatenation penalty compared to Client library test that uses hardcoded "100 Main Street". I would use the same approach, does not matter which one.

dzou: Done.

google-cloud-spanner-hibernate-performance-testing/src/main/java/com/google/cloud/spanner/hibernate/HibernateDriver.java
dzou: Done.

google-cloud-spanner-hibernate-performance-testing/src/main/java/com/google/cloud/spanner/hibernate/ClientLibraryOperations.java
dzou: Done; I decided to just surround this in a try/catch and suppress the error if it is a table-not-found error.

google-cloud-spanner-hibernate-performance-testing/src/main/java/com/google/cloud/spanner/hibernate/ClientLibraryOperations.java
```diff
@@ -72,9 +78,19 @@ public ClientLibraryOperations() {
   }
 
   /**
-   * Creates a database.
+   * Resets the test Spanner database.
    */
-  public void createTestDatabase() {
+  public void resetTestDatabase() {
+    try {
+      databaseAdminClient.dropDatabase(INSTANCE_NAME, DATABASE_NAME);
```
meltsufin: Is there a way to test if the table exists before attempting to drop it?

google-cloud-spanner-hibernate-performance-testing/src/main/java/com/google/cloud/spanner/hibernate/HibernateOperations.java
```diff
@@ -0,0 +1,94 @@
...
+final class HibernateOperations {
+
+  private StandardServiceRegistry registry;
+  private SessionFactory sessionFactory;
+
+  HibernateOperations() {
+    this.registry = new StandardServiceRegistryBuilder().configure().build();
+  }
+
+  void initializeEntityTables() {
```
meltsufin: I understand that this is test code, but it would be nice to have some consistency. All methods should probably be public, including the class itself.

google-cloud-spanner-hibernate-performance-testing/src/main/java/com/google/cloud/spanner/hibernate/HibernateOperations.java
```diff
@@ -0,0 +1,94 @@
...
+    this.sessionFactory = new MetadataSources(registry).buildMetadata().buildSessionFactory();
+  }
+
+  void insertRows(int rowCount) {
+    Session session = this.sessionFactory.openSession();
+    session.setFlushMode(FlushModeType.COMMIT);
```
meltsufin: Is this necessary? What's the purpose?
Also, it looks like you will be repeating this pattern of creating a session, starting a transaction, and then committing it a lot. So, it might make sense to create a `runInTransaction(lambda)` method for some reuse.

dzou: Done.

google-cloud-spanner-hibernate-performance-testing/src/main/java/com/google/cloud/spanner/hibernate/HibernateOperations.java
dzou: Done.

google-cloud-spanner-hibernate-performance-testing/src/main/java/com/google/cloud/spanner/hibernate/ClientLibraryOperations.java
dzou: The only way is to call databaseClient.listDatabases(..) which returns a paginated list and checking if the database is in there. I just decided this way is probably simpler and would only require 1 trip to the database.

PR #145, opened 2019-11-05
Cleanup client library update operation for performance testing

dzou:
Modify the client library update operation to accept a parameter for how many rows to update.


google-cloud-spanner-hibernate-performance-testing/src/main/java/com/google/cloud/spanner/hibernate/ClientLibraryOperations.java
```diff
@@ -169,13 +169,13 @@ public void batchInsert(int batchSize) {
...
    */
-  public void batchUpdate() {
+  public void batchUpdate(int count) {
     Statement statement = Statement.newBuilder("SELECT * FROM " + AIRPORT_TABLE).build();
     ResultSet resultSet =
         databaseClient.singleUseReadOnlyTransaction().executeQuery(statement);
 
     ArrayList<String> allRowIds = new ArrayList<>();
-    while (resultSet.next()) {
+    while (resultSet.next() && allRowIds.size() < count) {
```
meltsufin: Can you do a `LIMIT` query above instead to guarantee that there is no more than count ids?
You can also just select `id` rather than `*`.

dzou: SG, Done.

PR #151, opened 2019-11-13
Move setting the Bulk ID strategy to the service contributor

dzou:
This moves the setting of the Bulk ID strategy to the service contributor so users do not have to worry about setting it manually in configuration.


dzou:
> Great!
> I think we need to apply the same fix [here](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/spanner/hibernate/src/main/resources/hibernate.cfg.xml), but be careful there because that's used in the public docs as a snippet.

Sounds good, I created https://github.com/GoogleCloudPlatform/google-cloud-spanner-hibernate/issues/153 as a reminder.

PR #155, opened 2019-11-14
Added Quarkus JVM mode example

saturnism:


meltsufin:
Added #157 to track integration test.

PR #173, opened 2019-12-03
Add support for Unique Constraint in Hibernate

### LINKED ISSUE #165: Support @Column(unique = true) unique constraints

dzou:
This adds support for Hibernate to generate unique indices when uniqueness is invoked such as through annotating an entity column with `@Column(unique = true)`.

Fixes #165.


google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/schema/SpannerTableStatements.java
```diff
@@ -53,17 +57,38 @@ public SpannerTableStatements(SpannerDialect spannerDialect) {
...
+    Iterator<UniqueKey> keyIterator = table.getUniqueKeyIterator();
+    while (keyIterator.hasNext()) {
+      tableIndices.add(keyIterator.next().getName());
+    }
+
+    Iterator<Column> colIterator = table.getColumnIterator();
+    while (colIterator.hasNext()) {
+      Column col = colIterator.next();
+      if (col.isUnique()) {
+        tableIndices.add(Constraint.generateName("UK_", table, col));
```
elefeint: This won't show up in table.getUniqueKeyIterator() earlier? It was added in `table.getOrCreateUniqueKey(name)` during initialization.

google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/SpannerTableExporter.java
```diff
@@ -82,4 +88,19 @@ public void initializeTableExporter(Metadata metadata, Action schemaAction) {
...
+  /**
+   * Processes the columns of the table and creates Unique Constraints for columns
+   * annotated with @Column(unique = true).
+   */
+  private static void initializeUniqueConstraints(Table table) {
+    Iterator<Column> colIterator = table.getColumnIterator();
+    while (colIterator.hasNext()) {
+      Column col = colIterator.next();
+      if (col.isUnique()) {
+        String name = Constraint.generateName("UK_", table, col);
```
elefeint: Since "UK_" is used twice, consider making a constant. As a side benefit, I'll stop thinking about United Kingdom :D

google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/schema/SpannerTableStatements.java
```diff
@@ -53,17 +57,38 @@ public SpannerTableStatements(SpannerDialect spannerDialect) {
...
-      dropStrings.add("drop index " + curr.getName());
+    for (String indexName : getTableIndices(table)) {
+      dropStrings.add("drop index " + indexName);
     }
 
     dropStrings.add(this.spannerDialect.getDropTableString(table.getQuotedName()));
     return dropStrings;
   }
 
+  private Set<String> getTableIndices(Table table) {
```
meltsufin: Do we have test coverage for all 3 types of sources of indices?

meltsufin:
Are there any userguide tests we can re-enable now?

google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/SpannerTableExporter.java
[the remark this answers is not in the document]
dzou: "UK" appears to be the convention for naming unique constraints in Hibernate - I saw it named this way in the default Hibernate table exporter impl: https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/main/java/org/hibernate/tool/schema/internal/StandardTableExporter.java#L113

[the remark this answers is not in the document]
meltsufin: Alright then.

google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/schema/SpannerTableStatements.java
dzou: Nice catch, this is not needed. Removed.

google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/SpannerTableExporter.java
dzou: After removing the previous usage it will be used once so i'll just leave it inline.

google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/schema/SpannerTableStatements.java
dzou: Yes! After addressing Elena's comment there are only two sources of indices we need to address:
- "unique indices" - created by the `@Column(unique = true)`; the unit tests added verify this.
- "normal indices" - created by using `@Index(...)`. This is present on the `Employee` entity and is verified by pre-existing tests.

dzou:
> Are there any userguide tests we can re-enable now?

Yup - turns out the userguide test has a unique constraint test. This test was not written in a way that failed if the unique constraint was not enforced tho; so I modified it so that it correctly confirms an exception occurs if the unique constraint is broken on a DML operation: https://github.com/GoogleCloudPlatform/google-cloud-spanner-hibernate/blob/cc4a19dcbc6ab5e67dbda245a9efced2c4469c9e/google-cloud-spanner-hibernate-testing/src/test/java/org/hibernate/userguide/schema/UniqueConstraintTest.java

google-cloud-spanner-hibernate-testing/src/test/java/org/hibernate/userguide/schema/UniqueConstraintTest.java
```diff
@@ -55,17 +56,14 @@ public void test() {
...
-				Book book = new Book();
-				book.setTitle( "High-Performance Java Persistence" );
-				book.setAuthor( _author );
-				entityManager.persist( book );
-			} );
-        }
-        catch (Exception expected) {
-            assertNotNull( ExceptionUtil.findCause( expected, ConstraintViolationException.class ) );
-        }
+                Book book = new Book();
```
meltsufin: Should we contribute this fix back to Hibernate repo?

dzou: Sure, I can look into that.

PR #174, opened 2019-12-10
Implement batching statements in Hibernate UPDATE mode

### LINKED ISSUE #149: Implemented Batched DDL for the Schema Migrator (Update)
Implement batched DDL for the Schema Migrator.

The schema migrator is responsible for table schema updates. This implementation is provided by the `SchemaManagementTool`.

dzou:
Implement statement batching for Hibernate when running in Update mode.

Fixes #149.

### DIFF
--- a/google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/SpannerServiceContributor.java
+++ b/google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/SpannerServiceContributor.java
@@ -22,6 +22,7 @@ import com.google.cloud.spanner.hibernate.schema.SpannerSchemaManagementTool;
 import org.hibernate.boot.registry.StandardServiceRegistryBuilder;
 import org.hibernate.hql.spi.id.inline.InlineIdsOrClauseBulkIdStrategy;
 import org.hibernate.service.spi.ServiceContributor;
+import org.hibernate.tool.hbm2ddl.UniqueConstraintSchemaUpdateStrategy;
 
 /**
  * An implementation of a Hibernate {@link ServiceContributor} which provides custom settings
@@ -46,6 +47,10 @@ public class SpannerServiceContributor implements ServiceContributor {
         .applySetting("hibernate.connection.userAgent", HIBERNATE_API_CLIENT_LIB_TOKEN)
         // The custom Hibernate schema management tool for Spanner.
         .applySetting("hibernate.schema_management_tool", SCHEMA_MANAGEMENT_TOOL)
+        // Create a unique index for a table if it does not already exist when in UPDATE mode.
+        .applySetting(
+            "hibernate.schema_update.unique_constraint_strategy",
+            UniqueConstraintSchemaUpdateStrategy.RECREATE_QUIETLY)
         // Allows entities to be used with InheritanceType.JOINED in Spanner.
         .applySetting("hibernate.hql.bulk_id_strategy", InlineIdsOrClauseBulkIdStrategy.INSTANCE);
   }
--- a/google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/schema/RunBatchDdl.java
+++ b/google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/schema/RunBatchDdl.java
@@ -21,12 +21,19 @@ package com.google.cloud.spanner.hibernate.schema;
 import com.google.cloud.spanner.hibernate.SpannerDialect;
 import org.hibernate.boot.model.relational.AuxiliaryDatabaseObject;
 import org.hibernate.dialect.Dialect;
+import org.hibernate.tool.schema.Action;
 
 /**
  * Custom {@link AuxiliaryDatabaseObject} which generates the RUN BATCH statement.
  */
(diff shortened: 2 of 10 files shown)


google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/SpannerServiceContributor.java
```diff
@@ -46,6 +47,10 @@ public void contribute(StandardServiceRegistryBuilder serviceRegistryBuilder) {
         .applySetting("hibernate.connection.userAgent", HIBERNATE_API_CLIENT_LIB_TOKEN)
         // The custom Hibernate schema management tool for Spanner.
         .applySetting("hibernate.schema_management_tool", SCHEMA_MANAGEMENT_TOOL)
+        // Create a unique index for a table if it does not already exist when in UPDATE mode.
+        .applySetting(
+            "hibernate.schema_update.unique_constraint_strategy",
+            UniqueConstraintSchemaUpdateStrategy.RECREATE_QUIETLY)
```
meltsufin: Why do we need to stray from the default here?

google-cloud-spanner-hibernate-dialect/src/test/java/com/google/cloud/spanner/hibernate/GeneratedUpdateTableStatementsTests.java
```diff
@@ -0,0 +1,123 @@
...
+        "INSERT INTO hibernate_sequence (next_val) VALUES(1)",
+        "create index name_index on Employee (name)",
+        "RUN BATCH"
+    );
+  }
+
+  @Test
+  public void testUpdateStatements_alterTables() throws SQLException {
+    // This sets up which pre-existing tables that Hibernate sees.
+    mockConnection.setMetaData(MockJdbcUtils.createMockDatabaseMetaData("Employee"));
```
meltsufin: Can you refactor a bit so that the two tests don't have so much duplication?

google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/schema/RunBatchDdl.java
```diff
@@ -50,6 +61,10 @@ public boolean beforeTablesOnCreation() {
 
   @Override
   public String[] sqlDropStrings(Dialect dialect) {
-    return new String[] {"RUN BATCH"};
+    if (schemaAction == Action.UPDATE) {
+      return new String[]{};
```
meltsufin: Why `UPDATE` doesn't need `RUN BATCH`?

google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/schema/StartBatchDdl.java
```diff
@@ -40,7 +47,11 @@ public boolean appliesToDialect(Dialect dialect) {
 
   @Override
   public boolean beforeTablesOnCreation() {
-    return true;
+    if (schemaAction == Action.UPDATE) {
```
meltsufin: I see this logic repeated in several places. Can we refactor to have it in one place?

google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/SpannerServiceContributor.java
dzou: The default strategy is to drop a unique index and create it in Spanner; it will produce the drop statement regardless of whether it exists or not for the table. However it the unique index does not exist then that will produce an error for Spanner.

The `recreate_quietly` strategy omits the initial drop operation at the beginning.

google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/schema/RunBatchDdl.java
dzou: The AbstractSchemaMigrator triggers the [drop statement and then the create statement for each auxiliary object](https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/main/java/org/hibernate/tool/schema/internal/AbstractSchemaMigrator.java#L167), but this would cause us to produce the statements in this order:

```
RUN BATCH
START BATCH DDL
... all your ddl statements ...
RUN BATCH
```

So I make it so it does not produce a drop string so that it will not produce the extra RUN BATCH.

```
START BATCH DDL
... all your ddl statements ...
RUN BATCH
```

google-cloud-spanner-hibernate-dialect/src/test/java/com/google/cloud/spanner/hibernate/GeneratedUpdateTableStatementsTests.java
dzou: Done.

google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/schema/StartBatchDdl.java
dzou: Done.

PR #178, opened 2019-12-16
Check existence of tables/indices before Schema create/delete

### LINKED ISSUE #176: Batch schema generation causes the whole batch to fail if one table is missing
Maybe the solution is to just make the batching optional.
Another option is to query the database metadata and remove non-existing tables from the batch.

dzou:
This checks the existence of tables and indices before attempting to create or delete them in Spanner. This is necessary since dropping an nonexistent table or creating a table with a duplicate name will both result in error.

Fixes #176.

### DIFF
--- a/google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/schema/SpannerSchemaManagementTool.java
+++ b/google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/schema/SpannerSchemaManagementTool.java
@@ -19,7 +19,10 @@
 package com.google.cloud.spanner.hibernate.schema;
 
 import com.google.cloud.spanner.hibernate.SpannerTableExporter;
+import java.sql.DatabaseMetaData;
+import java.sql.SQLException;
 import java.util.Map;
+import org.hibernate.resource.transaction.spi.DdlTransactionIsolator;
 import org.hibernate.tool.schema.internal.HibernateSchemaManagementTool;
 import org.hibernate.tool.schema.internal.exec.JdbcContext;
 import org.hibernate.tool.schema.spi.ExecutionOptions;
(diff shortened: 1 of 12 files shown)


google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/schema/SpannerSchemaManagementTool.java
```diff
@@ -52,4 +55,15 @@ SpannerTableExporter getSpannerTableExporter(ExecutionOptions options) {
     JdbcContext jdbcContext = this.resolveJdbcContext(options.getConfigurationValues());
     return (SpannerTableExporter) jdbcContext.getDialect().getTableExporter();
   }
+
+  SpannerDatabaseInfo getDatabaseMetaData(ExecutionOptions options) {
```
meltsufin: How expensive is this operation? Does Hibernate already do something like this in the schema update mode? Can we re-use their logic?

google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/schema/SpannerTableStatements.java
```diff
@@ -47,20 +47,36 @@
...
   }
 
+  /**
+   * Initializes the {@link SpannerDatabaseInfo} which contains information about what tables and
+   * indices exist in the Spanner database.
+   *
+   * @param spannerDatabaseInfo the {@link SpannerDatabaseInfo} to load.
+   */
+  public void initializeSpannerDatabaseInfo(SpannerDatabaseInfo spannerDatabaseInfo) {
+    this.spannerDatabaseInfo = spannerDatabaseInfo;
```
meltsufin: This doesn't really initialize anything. Also, is there risk of the `SpannerDatabaseInfo` getting stale throughout the life of this object?

google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/schema/SpannerSchemaManagementTool.java
dzou: How expensive is this operation?
- It is not expensive, the added runtime is trivial. Hibernate does this on every update operation.

Does Hibernate already do something like this in the schema update mode?
- Yes, the code I got for doing this came out of Hibernate. However we have to retrieve the metadata info again, I was not able to find a way to retrieve the existing information.

Can we re-use their logic?
- Yup, this is copied over here, but I wasn't able to find a way to reuse the information they computed.

google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/schema/SpannerTableStatements.java
dzou: I prefer "initialize" just to be consistent how I used the term elsewhere - the idea just being that these are fields which could only be computed after the object was created and must be set now. Like you must "init" before you use the object.

No there is no risk because the initialize is called at the beginning of every create, drop, and update operation.

PR #183, opened 2020-01-07
Fix the Quarkus sample and add Integration Tests

dzou:
This fixes the dialect to be able to run the Quarkus sample correctly and adds integration tests to verify the sample.

### DIFF
--- a/google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/SpannerServiceContributor.java
+++ b/google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/SpannerServiceContributor.java
@@ -19,10 +19,15 @@
 package com.google.cloud.spanner.hibernate;
 
 import com.google.cloud.spanner.hibernate.schema.SpannerSchemaManagementTool;
+import java.util.Map;
+import org.hibernate.boot.registry.StandardServiceInitiator;
 import org.hibernate.boot.registry.StandardServiceRegistryBuilder;
 import org.hibernate.hql.spi.id.inline.InlineIdsOrClauseBulkIdStrategy;
+import org.hibernate.service.Service;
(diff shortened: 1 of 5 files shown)


google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/SpannerServiceContributor.java
```diff
@@ -43,10 +48,21 @@
...
+
+          @Override
+          public Class getServiceInitiated() {
+            return SchemaManagementTool.class;
+          }
+        })
         // The user agent JDBC connection property to identify the library.
         .applySetting("hibernate.connection.userAgent", HIBERNATE_API_CLIENT_LIB_TOKEN)
-        // The custom Hibernate schema management tool for Spanner.
-        .applySetting("hibernate.schema_management_tool", SCHEMA_MANAGEMENT_TOOL)
```
meltsufin: So, was this an incorrect way to specify the schema management tool, and initiator should be use instead?
What about the other `.applySetting()` calls we have? Do those also not work on Quarkus?
Will the `userAgent` be set?

dzou: > So, was this an incorrect way to specify the schema management tool, and initiator should be use instead?

Yeah, Sanne claimed setting properties is not expected to be done in the service contributor. The applySetting didn't work for changing the schema management tool in Quarkus and so I had to rely on the initiator method.

> What about the other .applySetting() calls we have? Do those also not work on Quarkus?
> Will the userAgent be set?

Good question, I will have to dig deeper into this. Right now I am not sure.

dzou: Filed #184 to track.

PR #204, opened 2020-09-15
Update best practices doc

dzou:
Update docs with Mike's discovery about `persist` versus `save`.

Adds a TOC.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -1,3 +1,6 @@
+:toc:
+:toclevels: 4
+
 = Google Cloud Spanner Dialect for Hibernate ORM
 
 This is a dialect compatible with https://hibernate.org/orm/releases/5.4/[Hibernate 5.4] for the https://cloud.google.com/spanner/[Google Cloud Spanner] database service.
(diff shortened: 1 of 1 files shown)


dzou:
Nevermind the wiki, just gonna leave on the homepage and add a TOC. Hope this looks nicer:

new view: https://github.com/GoogleCloudPlatform/google-cloud-spanner-hibernate/blob/53a6a281e2fb9cae0ec76b8ad11e10302cfaaf3a/README.adoc#quick-set-up

dmitry-s (dismissed):
LGTM

README.adoc
```diff
@@ -88,9 +91,13 @@ You can then activate the profile using `-Pspanner` such as `mvn compile -Pspann
 
 == User Guide
 
-The following section describes recommendations and best practices for how to best use the Cloud Spanner Hibernate Dialect.
+This guide contains a variety of best practices for using Hibernate with Spanner which can significantly improve the performance of your application.
+
+=== Schema Creation and Entity Design
 
-=== Use Generated UUIDs for ID Generation
+Hibernate generates statements based on your Hibernate entity's design. Following these practices can result in better DDL and DML statement generation which can improve performance.
```
elefeint, line 98:
```suggestion
Hibernate generates statements based on your Hibernate entity design. Following these practices can result in better DDL and DML statement generation which can improve performance.
```

elefeint: Because it's going to be based on multiple entities, and doing "entity's" vs "entities" is too pedantic for me. "entity design" covers both.

README.adoc
```diff
@@ -88,9 +91,13 @@ You can then activate the profile using `-Pspanner` such as `mvn compile -Pspann
...
+
+==== Use Generated UUIDs for ID Generation
```
elefeint, line 100: Do you want to comment on why autogenerated is better than self-created?

README.adoc
```diff
@@ -253,23 +211,95 @@ public class Album {
...
+```
 
-The following sections below describe the feature set of the Cloud Spanner Hibernate Dialect as well as its limitations.
+This practice may result in Hibernate executing additional `SELECT` statements that you may not have expected because it will perform a `SELECT` on the table to determine if a record in the table already has the entity's ID. If so, it will perform an update, otherwise it will save a new entity.
+
+These extra `SELECT` statements could yield large performance issues because they may lock an entire table depending on how they are generated.
+
+In these cases, we recommend to either:
+
+* Let Hibernate generate the ID by leaving the entity's `id` null and annotate the field with `@GeneratedValue`. Hibernate will know that the record did not exist prior if it generates a new ID.
```
elefeint, line 235: This might be better in "use generated UUIDs" section, with this current section just linking to it for the first solution. And then this section would have only the use-persist-to-be-explicit-about-inserts.

README.adoc
```diff
@@ -253,23 +211,95 @@ public class Album {
...
 [options="header"]
 |===
 | Unsupported Feature | Description
+| Large DML Transactions | Each Spanner transaction may only have up to 20,000 operations which modify rows of a table.
 | Catalog and schema scoping for table names | Tables name references cannot contain periods or other punctuation.
-| Column default values | The dialect does not currently set default values based on the `@ColumnDefault` annotation, because Cloud Spanner does not support column defaults in the DDL.
+| Column default values | The dialect does not set default values based on the `@ColumnDefault` annotation, because Cloud Spanner does not support column defaults in the DDL.
 |===
 
+=== Limit Large DML Transactions
```
elefeint, line 287: nit: other headings are noun phrases. Perhaps "Large DML transactions" or "Large DML transactions limits"?

README.adoc
```diff
@@ -253,23 +211,95 @@ public class Album {
...
+
+Cloud Spanner has a mutation limit on each transaction - each Spanner transaction https://cloud.google.com/spanner/quotas#limits_for_creating_reading_updating_and_deleting_data[may only have up to 20,000 operations which modify rows of a table].
+
+NOTE: Deleting a row counts as one operation and inserting/updating rows will count as a number of operations equal to the number of affected columns.
```
elefeint, line 291:
```suggestion
NOTE: Deleting a row counts as one operation and inserting/updating a single row will count as a number of operations equal to the number of affected columns.
```

elefeint: Over multiple rows, it would be numRows x numColumns, right? Might want to link to https://cloud.google.com/spanner/quotas#note2 or similar, since this is an intrinsically confusing aspect.

elefeint (changes requested):
Really useful to have this written down.

README.adoc
```diff
@@ -253,23 +211,95 @@ public class Album {
...
 
-The following sections below describe the feature set of the Cloud Spanner Hibernate Dialect as well as its limitations.
+This practice may result in Hibernate executing additional `SELECT` statements that you may not have expected because it will perform a `SELECT` on the table to determine if a record in the table already has the entity's ID. If so, it will perform an update, otherwise it will save a new entity.
+
+These extra `SELECT` statements could yield large performance issues because they may lock an entire table depending on how they are generated.
+
+In these cases, we recommend to either:
+
+* Let Hibernate generate the ID by leaving the entity's `id` null and annotate the field with `@GeneratedValue`. Hibernate will know that the record did not exist prior if it generates a new ID.
+* Or use `session.persist()` which explicitly will simply attempt the insert without generating an `ID`.
```
meltsufin, line 236:
```suggestion
* Or use `session.persist()` which will simply attempt the insert without generating an `ID`.
```

meltsufin (changes requested):
I think we should add a section about connection pooling.
You can base it off the discussion we are already having with Knut.

README.adoc
dzou: Done.

README.adoc
dzou: Done.

README.adoc
dzou: Done.

README.adoc
dzou: Done.

README.adoc
dzou: Done.

README.adoc
dzou: Done.

dzou:
> I think we should add a section about connection pooling.
> You can base it off the discussion we are already having with Knut.

Added references to the JDBC settings about GRPC channels. Regarding JDBC pooling - maybe we can add this in after experimenting with the JDBC pools; unclear from conversations how much performance you would gain from using them.

README.adoc
```diff
@@ -251,6 +241,17 @@ Batching can be enabled by configuring `hibernate.jdbc.batch_size` in your Hiber
 
 The property is set to `100` as an example; you may experiment with the batch size to see what works best for your application.
 
+==== JDBC Parameter Tuning
+
+The Spanner JDBC driver allows you to set the number of GRPC channels initialized through the JDBC connection URL.
+Each channel can support up to 100 concurrent requests; for application that require a high amount of concurrency this value can be increased (from the default of 4).
```
elefeint, line 247:
```suggestion
Each channel can support up to 100 concurrent requests; for applications that require a high amount of concurrency this value can be increased (from the default of 4).
```

README.adoc
```diff
@@ -279,16 +280,16 @@ The Cloud Spanner Hibernate Dialect supports most of the standard Hibernate and
 [options="header"]
 |===
 | Unsupported Feature | Description
-| Large DML Transactions | Each Spanner transaction may only have up to 20,000 operations which modify rows of a table.
+| Large DML Transaction Limits | Each Spanner transaction may only have up to 20,000 operations which modify rows of a table.
```
elefeint, line 283: This should probably stay as "Transaction", since the column header is "unsupported feature".

elefeint (changes requested):
Two tiny things.

README.adoc
```diff
@@ -126,56 +135,7 @@ public class Employee {
...
-                                         AND a.AlbumTitle LIKE '%love%';");
-
-// Executes the query.
-List<Object[]> entities = query.list();
-----
-
-Also, you may consult the https://cloud.google.com/spanner/docs/sql-best-practices[Cloud Spanner documentation] on general recommendations for optimizing performance.
-
-=== Schema Generation
+==== Generate Schema for faster development
```
meltsufin, line 138:
```suggestion
==== Auto-generate schema for faster development
```

README.adoc
```diff
@@ -190,7 +150,7 @@ The following settings are available:
 
 Hibernate performs schema updates on each table and entity type on startup, which can take more than several minutes if there are many tables. To avoid schema updates keeping Hibernate from starting for several minutes, you can update schemas separately and use the `none` or `validate` settings.
 
-==== Constraints Generation
+==== Use Constraints Generation to leverage Spanner foreign keys
```
meltsufin, line 153:
```suggestion
==== Leverage Cloud Spanner foreign key constraints
```

README.adoc
```diff
@@ -207,7 +167,7 @@ However, Cloud Spanner currently does not support cascading deletes on foreign k
 The dialect also supports unique column constraints applied through `@Column(unique = true)` or `@UniqueConstraint`.
 In these cases, the dialect will create a unique index to enforce uniqueness on the specified columns.
 
-==== Interleaved Tables
+==== Use Interleaved Tables for Parent-Child entities for better performance
```
meltsufin, line 170: inconsistent capitalization?
```suggestion
==== Use Interleaved Tables for parent-child relationships for better performance
```

This could also go under performance section, or link back to it?

README.adoc
```diff
@@ -253,23 +213,94 @@ public class Album {
 }
 ----
 
-== Cloud Spanner Hibernate ORM Limitations
+=== Performance Optimizations
+
+There are some practices which can improve the execution time of Hibernate operations.
+
+==== Be Clear about Inserts or Updates
```
meltsufin, line 220:
```suggestion
==== Be clear about inserts vs. updates
```

README.adoc
```diff
@@ -253,23 +213,94 @@ public class Album {
...
+
+Hibernate may generate additional `SELECT` statements if it is unclear whether you are attempting to insert a new record or update an existing record. The following practices can help with this:
+
+* Let Hibernate generate the ID by leaving the entity's `id` null and annotate the field with `@GeneratedValue`. Hibernate will know that the record did not exist prior if it generates a new ID. See the <<Use Generated UUIDs for ID Generation, above section>> for more details.
 
-The Cloud Spanner Hibernate Dialect supports most of the standard Hibernate and Java Persistence annotations, but there are important differences in features because of differences in https://cloud.google.com/spanner/docs/schema-and-data-model[Cloud Spanner's data model] from traditional SQL databases.
+* Or use `session.persist()` which will simply attempt the insert without generating an `ID`.
```
meltsufin, line 226: I believe it will still generate and ID if needed.
```suggestion
* Or use `session.persist()` which will explicitly attempt the insert.
```

meltsufin (changes requested):
We need to decide how we capitalize headings and be consistent.
Sentence case or Heading Case?

README.adoc
dzou: Done.

README.adoc
dzou: Done.

README.adoc
dzou: Done.

README.adoc
dzou: Done.

README.adoc
dzou: Done. Also moved to performance section.

README.adoc
dzou: Done.

README.adoc
dzou: Done.

dzou:
> We need to decide how we capitalize headings and be consistent.
> Sentence case or Heading Case?

I went through the headings and decided to just do heading case; it just looked a little more natural to me.

meltsufin (approved):
Heading case is good!

elefeint (approved):

Issue #194, opened 2020-06-10
Cannot build EntityManagerFactory for other database


ikenox:
I use `google-cloud-spanner-hibernate-dialect` v1.2.0 in my project, and I want to use also H2 Database in the same project for testing.

But, the following test code fails due to an error `Unable to build Hibernate SessionFactory`.

```java
    @Test
    public void test() {
        Map<String, Object> properties = Map.of(
            AvailableSettings.DRIVER, "org.h2.Driver",
            AvailableSettings.URL, "jdbc:h2:mem:test",
            AvailableSettings.DIALECT, "org.hibernate.dialect.H2Dialect",
            AvailableSettings.HBM2DDL_DATABASE_ACTION, "drop-and-create"
        );
        EntityManagerFactory test = Persistence.createEntityManagerFactory("test", properties);
    }
```

Stacktrace:
```
javax.persistence.PersistenceException: [PersistenceUnit: test] Unable to build Hibernate SessionFactory
        at org.hibernate.jpa.boot.internal.EntityManagerFactoryBuilderImpl.persistenceException(EntityManagerFactoryBuilderImpl.java:1336)
        at org.hibernate.jpa.boot.internal.EntityManagerFactoryBuilderImpl.build(EntityManagerFactoryBuilderImpl.java:1262)
        at org.hibernate.jpa.HibernatePersistenceProvider.createEntityManagerFactory(HibernatePersistenceProvider.java:56)
        at javax.persistence.Persistence.createEntityManagerFactory(Persistence.java:79)
        at com.example.H2Test.testSample(H2Test.java:19)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.base/java.lang.reflect.Method.invoke(Method.java:566)
        at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
        at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
        at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
[...]

PR #208, opened 2020-09-29
fix: do not add services if Spanner dialect is not used

olavloite:
The Spanner dialect would add services and configuration, even if the Spanner dialect was not used, but was present on the classpath. This makes it impossible to develop an application that can run on both Cloud Spanner and other databases,
as the Spanner services would be loaded and used for all databases.

It also made it impossible to use for example H2 for testing purposes, while using Cloud Spanner for production.

This PR also contains a change for the `SpannerServiceContributorTest`, as that test would not work on Java 11. This version should work on both Java 8 and Java 11.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,11 @@ target
 *.iml
 .idea
 
+# eclipse
+.project
+.settings
+.classpath
+
 # Ignores generated files through testing
 **/transaction.log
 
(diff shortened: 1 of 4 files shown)


olavloite:
CC @dzou

dzou:
Wow, thank you @olavloite !!!!

We did not realize you can conditionally trigger the serviceContributor. This fixes https://github.com/GoogleCloudPlatform/google-cloud-spanner-hibernate/issues/194 for us too. This is a really great fix.

I'll get this merged in and cut a new release shortly.

dzou:
cc/ @meltsufin

dzou (approved):

PR #218, opened 2020-10-07
fix: close session and sessionFactory when finished

olavloite:
The sample application should close the session and session factory when the application finishes. Failing to do so, will keep at least one JDBC connection open, which again could cause sessions to be kept around on the backend until they are garbage collected one hour later.

This chanage also turns off the cleanup of daemon threads by the maven-exec plugin. Otherwise it reports a large number of false positives of threads that are supposedly not shutdown. They are however closed by a shutdown hook after the check has been executed (the check is executed by this method: https://github.com/mojohaus/exec-maven-plugin/blob/d0e13ba1ddd383bc461f0efe547c91166ad789bf/src/main/java/org/codehaus/mojo/exec/ExecJavaMojo.java#L300)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,8 @@ target
 
 # local libs
 /libs
+
+# eclipse
+.project
+.settings
+.classpath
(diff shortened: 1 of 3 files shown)


olavloite:
cc @jjfox15

dzou:
@olavloite - Thank you for the contribution Knut and helping Sean look into the emulator issue. This is a great help; we really appreciate your time and expertise.

I'll get this merged in shortly.

dzou (approved):

PR #230, opened 2021-02-08
Fix Sonar no such element exception warning

dzou:
Fixes Sonar warning: https://sonarcloud.io/project/issues?id=google-cloud-spanner-hibernate&open=AXUn7HDWUbSbb-AErP2h&resolved=false&types=BUG


elefeint (changes requested):
Could you add a test for this?

dzou:
Going to close actually. I realized `hasNext()` has some side effects which complicates this. Will revisit later.

PR #232, opened 2021-03-09
Add Support for Spanner array column type

### LINKED ISSUE #231: Attempting to map a spanner array to a java array fails silently
When using EntityManager to try and run a native query, the query fails silently when trying to map array columns. 

Example:
```
em.createNativeQuery(
            "SELECT " +
                "tile.tile_id as id, " +
                "tile.slug, " +
                "details.display_name, " +
                "tile.seller_user_name, " +
                "tile.listing_source, " +
                "tile.size, " +
                "tile.color, " +
                "tile.brand, " +
                "tile.condition  " +
                "FROM " +
                "tile_detail as details " +
                "join " +
                "tile AS tile ON tile.tile_id = details.tile_id " +
                filter +
                " LIMIT :limit", CatalogListing::class.java
        )
            .setParameter("limit", totalItems)
            .resultList
```
 Fails when trying to get the resultList. 
 
 My data model (CatalogListing) That i am trying to map back to is this:
 
 ```
 @Entity
data class CatalogListing(
    @Id
    val id: String,
    val slug: String,
    val displayName: String,
    val sellerUserName: String,
    val listingSource: String,
    val size: String,
    val color: Array<String>,
    val brand: Array<String>,
    val condition: String
) 
```

If i remove the two array columns, the call works again. 

```
cloudSpannerHibernateDialectVersion=1.4.0
cloudSpannerJdbcVersion=1.17.0
```

### PR DESCRIPTION
**Pull request by dzou**
Maps `List<T>` to Spanner array column types `ARRAY<T>` in Hibernate entities.

Fixes #231.

### CODE REVIEW

File: google-cloud-spanner-hibernate-dialect/pom.xml [hunk 1]
```diff
@@ -18,5 +18,10 @@
       <groupId>org.hibernate</groupId>
       <artifactId>hibernate-core</artifactId>
     </dependency>
+
+    <dependency>
+      <groupId>com.google.cloud</groupId>
+      <artifactId>google-cloud-spanner-jdbc</artifactId>
```
meltsufin, line 24: How come we didn't need this before? I thought the idea was the that the user adds the jdbc dependency.

google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/reflection/ReflectionUtils.java
```diff
@@ -0,0 +1,78 @@
...
+import java.lang.reflect.Type;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.stream.Collectors;
+
+/**
+ * Provides helper functions to do Java Reflection.
+ */
+public class ReflectionUtils {
```
meltsufin, line 32: Would it make sense to just use Apache Commons?

google-cloud-spanner-hibernate-testing/src/main/java/com/google/cloud/spanner/hibernate/SpannerColumnTypeTest.java
```diff
@@ -0,0 +1,5 @@
+package com.google.cloud.spanner.hibernate;
+
+public class SpannerColumnTypeTest {
+
```
meltsufin, line 4: Were you planning to add something here?

meltsufin (commented):
Why can't we register the array type in the `SpannerDialect` constructor like we do for non-custom types?

google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/reflection/ReflectionUtils.java
dzou: Yeah makes sense -- I switched to using apache.commons.reflect3. Was able to remove this class.

google-cloud-spanner-hibernate-testing/src/main/java/com/google/cloud/spanner/hibernate/SpannerColumnTypeTest.java
dzou: My b, forgot I didn't actually add the test. Integration test added [here.](https://github.com/GoogleCloudPlatform/google-cloud-spanner-hibernate/pull/232/files#diff-82a61b6b278d60a08229a48a612fb37365a5fb97bb87bbd56e1c46c3cbbfb137)

google-cloud-spanner-hibernate-dialect/pom.xml
dzou: Yeah I only needed it to use the `Code` enum provided by the spanner libs. I'm using this enum in the [ArrayJavaTypeDescriptor](https://github.com/GoogleCloudPlatform/google-cloud-spanner-hibernate/pull/232/files#diff-7ffc33f6760659087bc9c552f081c3d9a90a5fe2be9e4fffdf3eaf96697630ffR140) to define the mapping between Java types and Spanner SQL codes.

I could also just map the Java type to a hardcoded String; but just thought using the Code enum would be cleaner. Tho if it's not worth having the dependency then I can use String.

Also something to note: I think the Spanner JDBC driver is much more mature than Simba, so users won't be choosing between the two. Maybe it makes sense to just include the spanner JDBC driver now.

dzou:
> Why can't we register the array type in the `SpannerDialect` constructor like we do for non-custom types?

Indeed, this is a good question. I thought this would be possible via [Dialect.contributeTypes(..)](https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/main/java/org/hibernate/dialect/Dialect.java#L371). You can actually register custom types there...

However, if you look deeper in the code, this only lets you register "Basic Types" via the `BasicTypeRegistry`. And for some reason, Hibernate does not inject [Dynamic parameters information into basic types](https://github.com/ekirkilevics/Hibernate/blob/master/hibernate-core/src/main/java/org/hibernate/cfg/annotations/SimpleValueBinder.java#L638). (This check `simpleValue.getMappings().getTypeResolver().basic( simpleValue.getTypeName() ) == null` blocks Hibernate from injecting the dynamic type info into the basic type)...

So I was not able to get this approach to work unfortunately. We need this dynamic information to be injected into our type so that we can know what the generic type of the List is; this informs what array type to send to JDBC driver.
The alternate "Type" you can register is called a `UserType` which is what we are doing with the `@TypeDefs` stuff. And indeed, this passes that condition since it is not in the basicTypeRegistry, so Hibernate injects that dynamic type infos into it.

google-cloud-spanner-hibernate-dialect/pom.xml
meltsufin: We only support the official driver anyway. So, I think it's fine. We can still leave the instructions to add the dependency to the app so that users can control the version independently.

meltsufin:
@dzou Can we just confirm with the Hibernate team about this limitation of the `BasicTypeRegistry`? Maybe they can even take a quick look over the approach.

meltsufin (approved):

google-cloud-spanner-hibernate-dialect/pom.xml
dzou: Got it.

dzou:
> @dzou Can we just confirm with the Hibernate team about this limitation of the `BasicTypeRegistry`? Maybe they can even take a quick look over the approach.

Sounds good. I posted a question here: https://hibernate.zulipchat.com/#narrow/stream/132096-hibernate-user/topic/Question.20about.20registering.20custom.20types.20in.20Dialect/near/230712133

Will wait a little for a response before merging this then.

dzou:
@meltsufin -- Got a response here: https://hibernate.zulipchat.com/#narrow/stream/132096-hibernate-user/topic/Question.20about.20registering.20custom.20types.20in.20Dialect/near/230712133

Seems like this is the only way -- Though apparently in Hibernate 6 they plan to make it built-in. :man_shrugging:

PR #233, opened 2021-03-10
Delete the `disable-integration-test` profile

dzou:
This removes the `disable-integration-test` profile, which should no longer be used after migrating to the fail-safe plugin which separates unit tests from integration tests. This profile was used to disable integration tests from being run by the surefire plugin back before I knew about differences between surefire vs. failsafe.

### DIFF
--- a/.kokoro/build.sh
+++ b/.kokoro/build.sh
@@ -20,7 +20,7 @@ set -eo pipefail
 dir=$(dirname "$0")
 
 pushd $dir/../
-./mvnw verify -B -V -P disable-integration-tests
+./mvnw install -B -V -DskipTests
 popd
 
 source $dir/release_snapshot.sh
--- a/.kokoro/build.sh
+++ b/.kokoro/build.sh
@@ -20,7 +20,7 @@ set -eo pipefail
 dir=$(dirname "$0")
 
 pushd $dir/../
-./mvnw install -B -V -DskipTests
+./mvnw install -B -V -DskipITs
 popd
 
 source $dir/release_snapshot.sh
--- a/.kokoro/stage.sh
+++ b/.kokoro/stage.sh
@@ -28,8 +28,8 @@ MAVEN_SETTINGS_FILE=$(realpath .)/settings.xml
 setup_environment_secrets
 create_settings_xml_file $MAVEN_SETTINGS_FILE
 
-# run unit tests
-./mvnw install -B -V
+# install and run unit tests
+./mvnw install -B -V -DskipITs
 
 # change to release version
 ./mvnw versions:set -DremoveSnapshot
--- a/.kokoro/stage.sh
+++ b/.kokoro/stage.sh
@@ -29,7 +29,7 @@ setup_environment_secrets
 create_settings_xml_file $MAVEN_SETTINGS_FILE
 
 # run unit tests
-./mvnw verify -B -V -P disable-integration-tests
+./mvnw install -B -V
 
 # change to release version
 ./mvnw versions:set -DremoveSnapshot
--- a/google-cloud-spanner-hibernate-samples/spring-data-jpa-sample/pom.xml
+++ b/google-cloud-spanner-hibernate-samples/spring-data-jpa-sample/pom.xml
@@ -79,12 +79,6 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-failsafe-plugin</artifactId>
-        <configuration>
-          <useModulePath>false</useModulePath>
-          <additionalClasspathElements>
-            <additionalClasspathElement>${basedir}/target/classes</additionalClasspathElement>
-          </additionalClasspathElements>
-        </configuration>
       </plugin>
     </plugins>
   </build>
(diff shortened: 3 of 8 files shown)


.kokoro/build.sh
```diff
@@ -20,7 +20,7 @@ set -eo pipefail
 dir=$(dirname "$0")
 
 pushd $dir/../
-./mvnw verify -B -V -P disable-integration-tests
+./mvnw install -B -V -DskipTests
```
meltsufin, line 23: Wouldn't this skip unit tests as well?

.kokoro/stage.sh
```diff
@@ -29,7 +29,7 @@ setup_environment_secrets
 create_settings_xml_file $MAVEN_SETTINGS_FILE
 
 # run unit tests
-./mvnw verify -B -V -P disable-integration-tests
+./mvnw install -B -V
```
meltsufin, line 32: `install` will run integration tests. Is that your intent here?

ttomsu:
If it helps, here's what I did in `spring-cloud-gcp`:

```
<properties>
  ...
    <skip.failsafe.tests>${skipTests}</skip.failsafe.tests>
    <skip.surefire.tests>${skipTests}</skip.surefire.tests>
    <skipTests>false</skipTests>
  ...
</properties>
```

```
<plugin>
  <groupId>org.apache.maven.plugins</groupId>
  <artifactId>maven-surefire-plugin</artifactId>
  <version>${maven-surefire-plugin.version}</version>
  <configuration>
    <argLine>-Xms512m -Xmx512m</argLine>
    <skip>${skip.surefire.tests}</skip>
    <excludes>
      <exclude>${integration-test.pattern}</exclude>
    </excludes>
  </configuration>
</plugin>
<plugin>
  <groupId>org.apache.maven.plugins</groupId>
  <artifactId>maven-failsafe-plugin</artifactId>
  <version>${maven-failsafe-plugin.version}</version>
  <configuration>
    <skip>${skip.failsafe.tests}</skip>
    <includes>
      <include>${integration-test.pattern}</include>
    </includes>
  </configuration>
</plugin>`
```

So in this case you case use `-DskipTests` to skip both kinds of tests but also `skip.failsafe.tests` or `skip.surefire.tests` to control them explicitly. 

In theory, I think there are properties already built in to control these, but in practice I found them not working as intuitively as I expected, so I added these.

.kokoro/stage.sh
dzou: Fixed, same as above.

dzou:
> If it helps, here's what I did in `spring-cloud-gcp`:
> 
> ```
> <properties>
>   ...
>     <skip.failsafe.tests>${skipTests}</skip.failsafe.tests>
>     <skip.surefire.tests>${skipTests}</skip.surefire.tests>
>     <skipTests>false</skipTests>
>   ...
> </properties>
> ```
> 
> ```
> <plugin>
>   <groupId>org.apache.maven.plugins</groupId>
>   <artifactId>maven-surefire-plugin</artifactId>
>   <version>${maven-surefire-plugin.version}</version>
>   <configuration>
>     <argLine>-Xms512m -Xmx512m</argLine>
>     <skip>${skip.surefire.tests}</skip>
>     <excludes>
>       <exclude>${integration-test.pattern}</exclude>
>     </excludes>
>   </configuration>
> </plugin>
> <plugin>
>   <groupId>org.apache.maven.plugins</groupId>
>   <artifactId>maven-failsafe-plugin</artifactId>
>   <version>${maven-failsafe-plugin.version}</version>
>   <configuration>
>     <skip>${skip.failsafe.tests}</skip>
>     <includes>
>       <include>${integration-test.pattern}</include>
>     </includes>
>   </configuration>
> </plugin>`
> ```
> 
> So in this case you case use `-DskipTests` to skip both kinds of tests but also `skip.failsafe.tests` or `skip.surefire.tests` to control them explicitly.
> 
> In theory, I think there are properties already built in to control these, but in practice I found them not working as intuitively as I expected, so I added these.

Thanks, this helps. I didn't realize there was a way to disable fail safe only. I'm gonna give using `-DskipITs` a try, this seems to be the property that disables failsafe. If this still runs into problems will try the property approach.

google-cloud-spanner-hibernate-samples/spring-data-jpa-sample/pom.xml
```diff
@@ -75,6 +75,11 @@
         <artifactId>spring-boot-maven-plugin</artifactId>
         <version>${spring-boot-version}</version>
       </plugin>
+
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-failsafe-plugin</artifactId>
```
meltsufin, line 81: Do you actually have to add this plugin? I thought it was built in.

meltsufin (approved):

google-cloud-spanner-hibernate-samples/spring-data-jpa-sample/pom.xml
dzou: Ah good catch. Removed.

PR #234, opened 2021-03-11
Migrate the CI to use Github Actions + Spanner Emulator

dzou:
Migrates the Hibernate CI to use Github Actions and the Spanner Emulator.

This will be a great improvement in the run time of the CI and the overall reliability (reduction of flakiness).

* Before, using Travis CI (26 mins + 5 mins queue time = ~30 mins): https://travis-ci.com/github/GoogleCloudPlatform/google-cloud-spanner-hibernate/builds/216112831
* Now, using GHA and emulator (~2 minutes): https://github.com/GoogleCloudPlatform/google-cloud-spanner-hibernate/runs/2100015775

There were 8/330 tests which fail under the emulator but not under the real Spanner instance; these were marked with a `@Ignore("Broken after emulator migration")`. I made [an issue](https://github.com/GoogleCloudPlatform/google-cloud-spanner-hibernate/issues/247) to look into them later. The error appears to be caused by when Hibernate tries to update the sequence table in the Emulator. I'm not sure why it fails. However I think the improvement in the CI is worth the temporary loss of the 8 tests.

Note also, it looks like in the PR there are big diffs in `BaseEntityManagerFunctionalTestCase.java` and `BaseCoreFunctionalTestCase.java`, but I just re-copied them over from the Hibernate repo. I wanted to reset these classes to what they originally were now that we are using the Spanner emulator. In the past we changed them significantly in order to make it so they don't create/delete tables after each test. Now with the emulator, creating/deleting tables per test is efficient and makes the tests less flaky (as we ensure the state is cleaned correctly each time).


google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/SpannerDialect.java
```diff
@@ -329,8 +329,8 @@ public String getCurrentSchemaCommand() {
 
   @Override
   public SchemaNameResolver getSchemaNameResolver() {
-    throw new UnsupportedOperationException(
-        "No schema name resolver supported by " + getClass().getName());
+    // Spanner does not have a notion of database name schemas, so return "".
```
elefeint, line 332: Throw the behavior change into release notes?

elefeint (approved):

google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/SpannerDialect.java
dzou: This change should be a no-op; I upgraded the version of hibernate in the PR and it seems like newer version requires us to return a schema resolver instead of throwing `UnsupportedOperationException()`. In either case though this object does not affect the user.

Issue #294, opened 2021-07-01
Dependabot config: Stay with Quarkus version 1.X


suztomo:
Quarkus 2 requires Java 11.  https://quarkus.io/blog/quarkus-2-0-0-final-released/

https://dependabot.com/docs/config-file/#ignored_updates

PR #298, opened 2021-07-07
Make dependabot ignore Quarkus 2

mpeddada1:
Quarkus 2.x version update PRs are still being created by dependabot. Follow up to #294.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,4 +9,4 @@ updates:
   - match:
       dependency_name: "io.quarkus:quarkus-universe-bom"
       # Quarkus 2 requires Java 11, which is too high.
-      version_requirement: ">=2.0.0.Final"
+      version_requirement: "2.x"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,4 +9,4 @@ updates:
   - match:
       dependency_name: "io.quarkus:quarkus-universe-bom"
       # Quarkus 2 requires Java 11, which is too high.
-      version_requirement: "2.x"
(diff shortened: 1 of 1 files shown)


suztomo (approved):

.github/dependabot.yml
```diff
@@ -9,4 +9,4 @@ updates:
   - match:
       dependency_name: "io.quarkus:quarkus-universe-bom"
       # Quarkus 2 requires Java 11, which is too high.
-      version_requirement: ">=2.0.0.Final"
+      version_requirement: "2.x"
```
suztomo, line 12: it's worth trying.

https://dependabot.com/docs/config-file/#ignored_updates says

> The range format is specific to the package manager

So, we can try `[2.0.0.Final,)` as per [this issue](https://github.com/dependabot/dependabot-core/issues/1845) if "2.x" does not work.

dzou (approved):

.github/dependabot.yml
mpeddada1: Nice find! As you mentioned, I will try "2.x" first and go with `[2.0.0.Final,)` if that doesn't work.

suztomo:
It seems working!

Another repository received dependabot PR for the 2.0.2.Final. This repository doesn't receive it. Good.
https://github.com/GoogleCloudPlatform/google-cloud-graalvm-support/pull/161

dzou:
Nice!

mpeddada1:
yay! Thanks for validating it.

elefeint:
:( #299. Let's see if @zhumin8 's comment puts it away for good.

PR #321, opened 2021-09-14
Add support for JSON column type

dzou:
Defines a custom Hibernate user type to support JSON column types.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -111,10 +111,20 @@ public class Employee {
 
 This results in sequential IDs that are https://cloud.google.com/spanner/docs/schema-and-data-model#primary_keys[not optimal for Cloud Spanner] and makes use of the `hibernate_sequence` table for generating IDs.
 
-==== Array Column Types
+==== Custom Spanner Column Types
 
-Spanner allows you to use https://cloud.google.com/spanner/docs/data-types#array_type[arrays as a data type] for your table columns.
-It is possible to map Hibernate entity fields to array columns by using the `com.google.cloud.spanner.hibernate.types.SpannerArrayListType`.
+Spanner offers custom column types not commonly found in other databases.
+This project offers https://docs.jboss.org/hibernate/orm/4.3/manual/en-US/html/ch06.html#types-custom[custom Hibernate type mappings] that correspond with these Spanner column types.
+
+|===
+| Spanner Data Type | Hibernate Type
+| https://cloud.google.com/spanner/docs/data-types#array_type[ARRAY]
+| `com.google.cloud.spanner.hibernate.types.SpannerArrayListType`
+| https://cloud.google.com/spanner/docs/data-types#json_type[JSON]
+| `com.google.cloud.spanner.hibernate.types.SpannerJsonType`
+|===
+
+You can use these type mappings through the Hibernate `@TypeDefs` system:
 
 [source, java]
 ----
--- a/README.adoc
+++ b/README.adoc
@@ -113,8 +113,7 @@ This results in sequential IDs that are https://cloud.google.com/spanner/docs/sc
 
 ==== Custom Spanner Column Types
 
-Spanner offers custom column types not commonly found in other databases.
-This project offers https://docs.jboss.org/hibernate/orm/4.3/manual/en-US/html/ch06.html#types-custom[custom Hibernate type mappings] that correspond with these Spanner column types.
+This project offers the following https://docs.jboss.org/hibernate/orm/4.3/manual/en-US/html/ch06.html#types-custom[custom Hibernate type mappings] for specific Spanner column types:
 
 [options="header"]
 |===
--- a/README.adoc
+++ b/README.adoc
@@ -116,6 +116,7 @@ This results in sequential IDs that are https://cloud.google.com/spanner/docs/sc
 Spanner offers custom column types not commonly found in other databases.
 This project offers https://docs.jboss.org/hibernate/orm/4.3/manual/en-US/html/ch06.html#types-custom[custom Hibernate type mappings] that correspond with these Spanner column types.
 
+[options="header"]
 |===
 | Spanner Data Type | Hibernate Type
 | https://cloud.google.com/spanner/docs/data-types#array_type[ARRAY]
--- a/google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/types/SpannerJsonType.java
+++ b/google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/types/SpannerJsonType.java
@@ -1,0 +1,47 @@
+/*
+ * Copyright 2019-2020 Google LLC
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA
+ */
+
+package com.google.cloud.spanner.hibernate.types;
+
+import com.google.cloud.spanner.hibernate.types.internal.JsonJavaTypeDescriptor;
+import com.google.cloud.spanner.hibernate.types.internal.JsonSqlTypeDescriptor;
+import java.util.Properties;
+import org.hibernate.type.AbstractSingleColumnStandardBasicType;
+import org.hibernate.usertype.DynamicParameterizedType;
+
+/**
+ * A {@link org.hibernate.usertype.UserType} which supports mapping Spanner JSON columns to
+ * entity fields.
+ */
+public class SpannerJsonType extends AbstractSingleColumnStandardBasicType<Object>
+    implements DynamicParameterizedType {
+
+  public SpannerJsonType() {
+    super(new JsonSqlTypeDescriptor(), new JsonJavaTypeDescriptor());
+  }
+
+  @Override
+  public String getName() {
+    return "spanner-json-type";
+  }
+
+  @Override
+  public void setParameterValues(Properties parameters) {
+    ((JsonJavaTypeDescriptor) getJavaTypeDescriptor()).setParameterValues(parameters);
+  }
+}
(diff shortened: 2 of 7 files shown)


google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/types/SpannerJsonType.java
```diff
@@ -0,0 +1,47 @@
...
+  @Override
+  public String getName() {
+    return "spanner-json-type";
+  }
+
+  @Override
+  public void setParameterValues(Properties parameters) {
+    ((JsonJavaTypeDescriptor) getJavaTypeDescriptor()).setParameterValues(parameters);
+  }
+}
```
meltsufin, line 47: Do you also need to override `registerUnderJavaType()` like they do here: https://github.com/vladmihalcea/hibernate-types/blob/master/hibernate-types-55/src/main/java/com/vladmihalcea/hibernate/type/json/JsonStringType.java?

google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/types/internal/JsonJavaTypeDescriptor.java
```diff
@@ -0,0 +1,75 @@
...
+import org.hibernate.type.descriptor.java.AbstractTypeDescriptor;
+import org.hibernate.usertype.DynamicParameterizedType;
+
+/**
+ * A {@link AbstractTypeDescriptor} for Spanner JSON columns.
+ */
+public class JsonJavaTypeDescriptor extends AbstractTypeDescriptor<Object>
+    implements DynamicParameterizedType {
+
+  private static final Gson gson = new Gson();
```
meltsufin, line 33: What's the reason for using Gson? Hibernate uses Jackson. (See: https://github.com/vladmihalcea/hibernate-types/blob/master/hibernate-types-55/src/main/java/com/vladmihalcea/hibernate/type/util/ObjectMapperWrapper.java)

google-cloud-spanner-hibernate-testing/src/test/java/org/hibernate/dialect/entities/JsonEntity.java
```diff
@@ -0,0 +1,127 @@
...
+import javax.persistence.GenerationType;
+import javax.persistence.Id;
+import org.hibernate.annotations.Type;
+import org.hibernate.annotations.TypeDef;
+import org.hibernate.annotations.TypeDefs;
+
+/**
+ * A Test entity with JSON entity fields.
+ */
+@TypeDefs({
```
meltsufin, line 17: Why do we need this `@TypeDef` and `hibernate-types` extension doesn't?
See: https://github.com/vladmihalcea/hibernate-types/blob/master/hibernate-types-55/src/test/java/com/vladmihalcea/hibernate/type/json/MySQLGenericJsonTypeTest.java#L79.

meltsufin (commented):
Have you seen the `hibernate-types` project that implements the same thing for a bunch of databases?
https://github.com/vladmihalcea/hibernate-types

Can the [`JsonStringType`](https://github.com/vladmihalcea/hibernate-types/blob/master/hibernate-types-55/src/main/java/com/vladmihalcea/hibernate/type/json/JsonStringType.java) work for us basically verbatim?

google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/types/internal/JsonJavaTypeDescriptor.java
elefeint: We looked at this with @dzou and @zhumin8 -- gson is brought in through the client library, so it's already present.

google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/types/internal/JsonJavaTypeDescriptor.java
```diff
@@ -0,0 +1,75 @@
...
+
+  // The JSON type declared in the entity.
+  private Class<?> propertyType = Object.class;
+
+  public JsonJavaTypeDescriptor() {
+    // This cast is needed to pass Object.class to the parent class
```
elefeint, line 39: What cast?

google-cloud-spanner-hibernate-testing/src/test/java/org/hibernate/dialect/entities/JsonEntity.java
elefeint: They seem to register types [programmatically](https://github.com/vladmihalcea/hibernate-types/blob/4ad5cec2f3539a1c5e95103849c215799b967d5f/hibernate-types-52/src/test/java/com/vladmihalcea/hibernate/type/util/AbstractTest.java#L251) for [MySQL](https://github.com/vladmihalcea/hibernate-types/blob/1b0fb16c03c4685488f6b9c67f0843c07117fdcb/hibernate-types-52/src/test/java/com/vladmihalcea/hibernate/type/json/MySQLParametrizedJsonTypeTest.java#L36) instead of with annotations.

There are [other types](https://github.com/vladmihalcea/hibernate-types/blob/6020ac94c530ac75b1ab70b7e2b0b28a5ee7f99f/hibernate-types-52/src/test/java/com/vladmihalcea/hibernate/type/json/PostgreSQLJsonMapTest.java#L63) where the declarative TypeDef approach is used.

elefeint (approved):

dzou:
> Have you seen the `hibernate-types` project that implements the same thing for a bunch of databases?
> https://github.com/vladmihalcea/hibernate-types
> 
> Can the [`JsonStringType`](https://github.com/vladmihalcea/hibernate-types/blob/master/hibernate-types-55/src/main/java/com/vladmihalcea/hibernate/type/json/JsonStringType.java) work for us basically verbatim?

Yes! I based this implementation off the hibernate-types project actually.

Nah not quite, there is a little bit of Spanner-specific code in our work that makes it not possible to re-use hibernate types verbatim. Some examples include:

- We have to use the correct JSON type code used by the JDBC driver, which is `JsonType.VENDOR_TYPE_NUMBER`. See: https://github.com/googleapis/java-spanner-jdbc/pull/447#issuecomment-828358151

- We must use `Value.json(..)` to pass a value to the JDBC driver, also described [here](https://github.com/googleapis/java-spanner-jdbc/pull/447#issuecomment-828358151). `Value.json(..)` is a class provided by Spanner client libraries.

Also, I would say that our way of doing things is slightly simpler than hibernate-types since we just have to support Spanner and not handle any special cases among other databases.

google-cloud-spanner-hibernate-testing/src/test/java/org/hibernate/dialect/entities/JsonEntity.java
meltsufin: Thanks for the clarification. I also now found that `@TypeDef` is required according to [this article](https://vladmihalcea.com/how-to-map-json-objects-using-generic-hibernate-types/).

meltsufin (approved):
There are still some feedback from me and @elefeint that you might optionally address. Otherwise, LGTM.

google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/types/SpannerJsonType.java
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * A {@link org.hibernate.usertype.UserType} which supports mapping Spanner JSON columns to
+ * entity fields.
+ */
+public class SpannerJsonType extends AbstractSingleColumnStandardBasicType<Object>
+    implements DynamicParameterizedType {
+
+  public SpannerJsonType() {
+    super(new JsonSqlTypeDescriptor(), new JsonJavaTypeDescriptor());
```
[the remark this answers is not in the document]
dzou: Actually I can't quite do this. So a `SpannerJsonType` gets created for each field on an entity that receives the JSON type annotation. However, this makes it so we cannot make `JsonJavaTypeDescriptor` static, since it carries state about what was the class type in the entity that received the annotation: https://github.com/GoogleCloudPlatform/google-cloud-spanner-hibernate/pull/321/files#diff-098729681e5bbcfe85a1db02a5402cccf0824e4683deed0e1231e65d6a385d8d
And it is this type that we convert from JSON string to object with using GSON.

I could theoretically make the `JsonSqlTypeDescriptor` static though since it does not carry state, but I think it would just be confusing to have one static and one not static, and lead to bugs if it needs to carry state in the future.

Hibernate-types also makes separate instances: https://github.com/vladmihalcea/hibernate-types/blob/master/hibernate-types-55/src/main/java/com/vladmihalcea/hibernate/type/json/JsonType.java#L42

google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/types/SpannerJsonType.java
dzou: I don't think so. Their [JsonType](https://github.com/vladmihalcea/hibernate-types/blob/master/hibernate-types-55/src/main/java/com/vladmihalcea/hibernate/type/json/JsonType.java#L42) does not override this method. It is not clear what effect the method has.

google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/types/SpannerJsonType.java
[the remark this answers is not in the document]
meltsufin: Fair enough, I'm not sure where I saw them make these static.

README.adoc
```diff
@@ -111,10 +111,21 @@ public class Employee {
 
 This results in sequential IDs that are https://cloud.google.com/spanner/docs/schema-and-data-model#primary_keys[not optimal for Cloud Spanner] and makes use of the `hibernate_sequence` table for generating IDs.
 
-==== Array Column Types
+==== Custom Spanner Column Types
 
-Spanner allows you to use https://cloud.google.com/spanner/docs/data-types#array_type[arrays as a data type] for your table columns.
-It is possible to map Hibernate entity fields to array columns by using the `com.google.cloud.spanner.hibernate.types.SpannerArrayListType`.
+Spanner offers custom column types not commonly found in other databases.
```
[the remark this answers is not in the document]
dzou: Done.

google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/types/internal/JsonJavaTypeDescriptor.java
dzou: Removed!

PR #336, opened 2021-11-01
Split tests into nightly and presubmit

dzou:
Nightly tests will be run each day and presubmit tests will be run as a check to the PR.

The motivation for this change is to allow a smaller set of tests to be run on presubmit against the emulator while running all tests against a real Spanner instance nightly. This is because the Spanner emulator does not have feature parity with the latest changes in Spanner.


suztomo:
Does this allow us to adopt newer version of the spanner jdbc library? if the tests with the emulator runs in the nightly test, then upgrading the library to the latest seems to fail the nightly builds.

dzou:
> Does this allow us to adopt newer version of the spanner jdbc library? if the tests with the emulator runs in the nightly test, then upgrading the library to the latest seems to fail the nightly builds.

Yeah, this will let us upgrade the JDBC version. I think for tests that pass running on the emulator, we can still keep as a presubmit check, while running all integration tests and sample tests as a nightly tests against Spanner. Still trying to work it out.

suztomo:
Now I got the idea. The nightly test will use real Spanner and this PR moves the problematic test cases there. Thank you.

elefeint (commented):
Why not run all tests on PRs (while still separating them into an emulator and IT phases)? How long do they take to run?

If we have some tests running nightly but not on PRs, eventually we'll get into a broken build situation as a result of PR, unless we are disciplined about running integration tests locally. We do it for external contributions in Spring Cloud GCP, but it's easy to forget.

dzou:
> Why not run all tests on PRs (while still separating them into an emulator and IT phases)? How long do they take to run?
> 
> If we have some tests running nightly but not on PRs, eventually we'll get into a broken build situation as a result of PR, unless we are disciplined about running integration tests locally. We do it for external contributions in Spring Cloud GCP, but it's easy to forget.

I see. They take 2-3 hours to run. I guess keeping as presubmit is not so bad for this repo since the project activity is low.

elefeint:
Yeah, with 2-3 hours I can see why you made them nightly. Withdrawing my comment.

dzou:
> Yeah, with 2-3 hours I can see why you made them nightly. Withdrawing my comment.

No problem. Lets just keep as nightly for now; I also have concerns that running at presubmit could also trigger concurrent builds which would be problematic for the tests. I think nightly would be good enough for our purposes.

elefeint (approved):

PR #424, opened 2022-06-06
Fixing dependabot.yml

ddixit14:


.github/dependabot.yml
```diff
@@ -8,8 +8,7 @@ updates:
...
     prefix: "deps:"
-  ignored_updates:
-  - match:
-      dependency_name: "io.quarkus:quarkus-universe-bom"
-      # Quarkus 2 requires Java 11, which is too high.
-      version_requirement: "2.x"
+  ignore:
+    - dependency-name: "io.quarkus:quarkus-universe-bom"
+    # Quarkus 2 requires Java 11, which is too high.
+    versions: "2.x"
```
elefeint, line 14: I think you might have to use the square brackets for a yaml array -- `["2.x"]`

elefeint (approved):

.github/dependabot.yml
zhumin8, line 14: Not sure how strict this is, but you might need extra indentation here. 
Also, the same dependency is also ignored in #299.

PR #441, opened 2022-06-23
temporary rollback of spring-boot-version

ddixit14:


elefeint:
What about this? https://github.com/GoogleCloudPlatform/google-cloud-spanner-hibernate/pull/438

ddixit14:
Okay I think it is already working, don't need this so closing it.

PR #443, opened 2022-06-24
Fixing breaking changes for microprofile major version bump

ddixit14:
This includes fixes for breaking changes due to microprofile major version bump from 4.1 to 5.0. The incompatibility were mainly due to namespace change from javax to jakarta in the code.


elefeint:
Nice! You can actually push those changes straight into dependabot branch -- `dependabot/maven/google-cloud-spanner-hibernate-samples/org.eclipse.microprofile-microprofile-5.0`.

PR #454, opened 2022-07-07
chore: Changes for release build job

ddixit14:


ddixit14:
> Why remove the release profile? It prevents samples and test project from getting released.
> 
> https://github.com/GoogleCloudPlatform/google-cloud-spanner-hibernate/blob/master/pom.xml#L304-L306

Sorry, it was by mistake. I actually intended to include the profile in the promote release part of the script, since the modules being processed in both the staging and releasing scripts were different.

elefeint (changes requested):
Do you need to specify the staging repository? In the old process, stage.sh sent all releasable artifacts to Sonatype, and then promote.sh triggered the publishing of those artifacts. This was all tied together by the "staging repository ID" -- the thing that you'd find in the staging logs and provide to the release/promote job.

The current [promote.sh](https://github.com/GoogleCloudPlatform/google-cloud-spanner-hibernate/blob/fabbc81e3924b1f12658760c5bf3e25eb0ecba4d/.kokoro/promote.sh#L39) passes `-DstagingRepositoryId=${STAGING_REPOSITORY_ID}` to let sonatype know which pre-staged directory to publish. What is the new process meant to do -- release in one step, or stage-and-release like before just with a single script?

.kokoro/stage.sh
```diff
@@ -46,9 +47,10 @@ create_settings_xml_file $MAVEN_SETTINGS_FILE
 # promote release
 if [[ -n "${AUTORELEASE_PR}" ]]
 then
-  mvn nexus-staging:release -B \
+  ./mvnw nexus-staging:release -B \
     -DperformRelease=true \
-    --settings=settings.xml
+    --settings=settings.xml \
+    -P release
```
suztomo, line 53: The missing problem. The 2nd command was trying to release the integration test module.

suztomo (approved):
Approving. Don't forget to answer Elena's question.

elefeint:
We've looked at it offline, and it looks like `mvn deploy` stages it first locally, and then remotely at Sonatype, generating the staging repository ID along the way. Then `mvn nexus-release` takes the already-staged files and the (cached?) staging repository ID and allows promotion of staged artifacts to released ones.

ddixit14:
@elefeint can you please approve this so I can merge it in?

pom.xml
```diff
@@ -175,7 +175,7 @@
           <version>1.6.13</version>
           <extensions>true</extensions>
           <configuration>
-            <serverId>ossrh</serverId>
+            <serverId>sonatype-nexus-staging</serverId>
```
elefeint, line 178: I'd keep `ossrh` in all three instances, unless there is a reason to change repository ID.

elefeint (changes requested):

elefeint (approved):

PR #471, opened 2022-10-17
fix: do not ignore exception during schema migration

### LINKED ISSUE #470: SQLExceptions during schema migrations are silently ignored
The `SpannerSchemaManagementTool` will automatically include the `START BATCH DDL` and `RUN BATCH` statements as auxiliary database objects for each schema migration. This ensures that each migration is executed as a single batch, which improves execution performance, and ensures that the parsing of the entire batch is atomic. This means that if one statement is invalid, the entire batch will be refused and nothing will be executed.

However, Hibernate will [ignore](https://github.com/hibernate/hibernate-orm/blob/0baefce734bed664a20f994ad35b0e0017aca770/hibernate-core/src/main/java/org/hibernate/tool/schema/internal/AbstractSchemaMigrator.java#L289) any `SQLException` that is thrown when executing the create/drop statement for an auxiliary database object. This means that if `RUN BATCH` fails, for example because one of the statements in the batch is invalid, the entire batch will not be executed, but Hibernate will not report any problems.

olavloite:
The Cloud Spanner Hibernate dialect will include the statements `START BATCH DDL` and `RUN BATCH` as auxiliary database objects. This ensures that the entire schema migration is executed as a single DDL batch, which both improves the execution speed of the statements, as well as makes the parsing of all statements atomic. However, Hibernate will ignore any SQLException that is thrown when creating an auxiliary database object. This means that if the `RUN BATCH` statement returns a SQLException, for example because one of the DDL statements is invalid, Hibernate will just silently ignore the error and proceed as if the schema migration worked.

This is fixed in this PR by returning an exception that is not an instance of `SQLException` specifically when one of the statements `START BATCH DDL` or `RUN BATCH` is executed as part of a schema migration.

Fixes #470

### DIFF
--- a/google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/schema/SpannerSchemaManagementTool.java
+++ b/google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/schema/SpannerSchemaManagementTool.java
@@ -18,8 +18,19 @@
 
 package com.google.cloud.spanner.hibernate.schema;
 
+import com.google.cloud.spanner.Dialect;
+import com.google.cloud.spanner.SpannerExceptionFactory;
+import com.google.cloud.spanner.connection.AbstractStatementParser;
+import com.google.cloud.spanner.connection.AbstractStatementParser.ParsedStatement;
+import com.google.cloud.spanner.connection.AbstractStatementParser.StatementType;
+import com.google.cloud.spanner.connection.StatementResult.ClientSideStatementType;
 import com.google.cloud.spanner.hibernate.SpannerTableExporter;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.lang.reflect.Proxy;
 import java.sql.Connection;
+import java.sql.SQLException;
+import java.sql.Statement;
 import java.util.Map;
 import org.hibernate.resource.transaction.spi.DdlTransactionIsolator;
 import org.hibernate.tool.schema.internal.HibernateSchemaManagementTool;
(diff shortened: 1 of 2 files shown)


google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/schema/SpannerSchemaManagementTool.java
```diff
@@ -34,6 +45,111 @@
...
+                            && args1 != null
+                            && args1.length == 1
+                            && args1[0] instanceof String) {
+                          // Check if the statement that is being executed is either `START BATCH
+                          // DDL` or `RUN BATCH`.
+                          String sql = (String) args1[0];
+                          ParsedStatement statement =
+                              PARSER.parse(com.google.cloud.spanner.Statement.of(sql));
+                          if (statement.getType() == StatementType.CLIENT_SIDE) {
+                            if (statement.getClientSideStatementType()
```
JoeWang1127, line 115: Could you combine two `if` statements in line 114-115?

google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/schema/SpannerSchemaManagementTool.java
```diff
@@ -34,6 +45,111 @@
...
+              delegateConnection.getClass().getClassLoader(),
+              new Class[] {Connection.class},
+              (proxy, method, args) -> {
+                // Only handle the Connection#createStatement() differently.
+                // All other methods are just passed through.
+                if (method.equals(createStatementMethod)) {
+                  // Create a proxy for the returned Statement that will override the behavior of
+                  // Statement#execute(String).
+                  Statement delegateStatement = delegateConnection.createStatement();
+                  return Proxy.newProxyInstance(
```
JoeWang1127, line 99: Do you think it's better to extract the `return` statement to a separate method?

google-cloud-spanner-hibernate-dialect/src/test/java/com/google/cloud/spanner/hibernate/GeneratedCreateTableStatementsTests.java
```diff
@@ -245,4 +248,19 @@ public void testCreateUniqueIndexes_oneToMany() throws SQLException {
...
+    this.connection
+        .getStatementResultSetHandler()
+        .prepareThrowsSQLException("RUN BATCH", new SQLException("test exception"));
+    Metadata metadata =
+        new MetadataSources(this.registry).addAnnotatedClass(Account.class).buildMetadata();
+
+    // Without the custom DdlTransactionIsolater that is returned by SpannerSchemaManagementTool,
+    // the SQLException would just be silently ignored by Hibernate.
+    SpannerException spannerException =
+        assertThrows(SpannerException.class, () -> metadata.buildSessionFactory().openSession());
```
JoeWang1127, line 263: Please separate the chained method so that only one invocation should throw an exception.

google-cloud-spanner-hibernate-dialect/src/test/java/com/google/cloud/spanner/hibernate/GeneratedCreateTableStatementsTests.java
```diff
@@ -245,4 +248,19 @@ public void testCreateUniqueIndexes_oneToMany() throws SQLException {
         "RUN BATCH"
     );
   }
+
+  @Test
+  public void testRunBatchFails() {
```
JoeWang1127, line 253: Could you also write a test for `START BATCH DDL`, or make this one a parameterized test.

google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/schema/SpannerSchemaManagementTool.java
olavloite: Done

google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/schema/SpannerSchemaManagementTool.java
olavloite: Yes, it will make it more readable. Done.

google-cloud-spanner-hibernate-dialect/src/test/java/com/google/cloud/spanner/hibernate/GeneratedCreateTableStatementsTests.java
olavloite: Done.

google-cloud-spanner-hibernate-dialect/src/test/java/com/google/cloud/spanner/hibernate/GeneratedCreateTableStatementsTests.java
JoeWang1127: Sure.

JoeWang1127 (approved):
LGTM

PR #478, opened 2022-10-27
feat: support bit-reversed sequences

olavloite:
Adds support for bit-reversed sequences. These are recommended above normal, monotonically increasing sequences for Cloud Spanner, as using a monotonically increasing value for a primary key can cause write hot spots.

This change also fixes the initialization of the backing tables for these sequences. Normally, the Cloud Spanner Hibernate dialect would not insert a value into the table that is backing a sequence, unless it was the default `hibernate_sequence` table. This would mean that generators that used custom table would be generated, but initialized, which would mean that they would not work out of the box.

### DIFF
--- a/google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/BitReversedSequenceStyleGenerator.java
+++ b/google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/BitReversedSequenceStyleGenerator.java
@@ -18,6 +18,7 @@
 
 package com.google.cloud.spanner.hibernate;
 
+import com.google.common.annotations.VisibleForTesting;
 import java.io.Serializable;
 import java.util.Properties;
 import org.hibernate.HibernateException;
@@ -160,10 +161,16 @@ public class BitReversedSequenceStyleGenerator extends SequenceStyleGenerator {
   @Override
   public Serializable generate(SharedSessionContractImplementor session, Object object)
       throws HibernateException {
-    Serializable id = super.generate(session, object);
+    Serializable id = generateBaseValue(session, object);
     if (id instanceof Long) {
       return Long.reverse((Long) id);
     }
     return id;
+  }
+
+  @VisibleForTesting
+  protected Serializable generateBaseValue(
+      SharedSessionContractImplementor session, Object object) {
+    return super.generate(session, object);
   }
 }
--- a/google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/BitReversedSequenceStyleGenerator.java
+++ b/google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/BitReversedSequenceStyleGenerator.java
@@ -1,0 +1,169 @@
+/*
+ * Copyright 2019-2020 Google LLC
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA
+ */
+
+package com.google.cloud.spanner.hibernate;
+
+import java.io.Serializable;
+import java.util.Properties;
+import org.hibernate.HibernateException;
+import org.hibernate.boot.model.naming.Identifier;
+import org.hibernate.boot.model.relational.Database;
+import org.hibernate.boot.model.relational.InitCommand;
+import org.hibernate.boot.model.relational.Namespace;
+import org.hibernate.boot.model.relational.QualifiedName;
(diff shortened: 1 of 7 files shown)


google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/SpannerTableExporter.java
```diff
@@ -55,47 +60,90 @@ public SpannerTableExporter(SpannerDialect spannerDialect) {
...
+              }
+            })
         .collect(Collectors.toList());
+  }
+
+  static void addStatementAfterDdlBatch(Metadata metadata, String[] statements) {
+    // Find the RunBatchDdl auxiliary object which can run statements after the DDL batch.
+    Optional<RunBatchDdl> runBatchDdl =
+        metadata.getDatabase().getAuxiliaryDatabaseObjects().stream()
+            .filter(obj -> obj instanceof RunBatchDdl)
```
JoeWang1127, line 127: Just a recommendation: you can replace lambda method reference here: `RunBatchDdl.class::isInstance`.

google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/BitReversedSequenceStyleGenerator.java
```diff
@@ -0,0 +1,169 @@
...
+        incrementSize,
+        type.getReturnedClass());
+  }
+
+  /**
+   * Generates a new ID. This uses the normal sequence strategy, but the returned ID is bit-reversed
+   * before it is returned to the application.
+   */
+  @Override
+  public Serializable generate(SharedSessionContractImplementor session, Object object)
```
JoeWang1127, line 161: Can you add a test to check the sequence is actually reversed?

google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/SpannerTableExporter.java
```diff
@@ -55,47 +60,90 @@ public SpannerTableExporter(SpannerDialect spannerDialect) {
...
+            .map(obj -> (RunBatchDdl) obj)
+            .findFirst();
 
-    return ddlStatements.toArray(new String[ddlStatements.size()]);
+    if (runBatchDdl.isPresent()) {
+      for (String statement : statements) {
+        runBatchDdl.get().addAfterDdlStatement(statement);
+      }
+    } else {
+      throw new IllegalStateException(
```
JoeWang1127, line 136: Can you add a test to throw an exception here?

google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/BitReversedSequenceStyleGenerator.java
```diff
@@ -0,0 +1,169 @@
...
+      throws HibernateException {
+    Serializable id = super.generate(session, object);
```
bharadwaj-aditya, line 163: From what value is the sequence initiated when the application starts up ?

olavloite: Depending on the initial state of the underlying database:
1. The sequence (table) does not yet exist: A new table is created and a row is inserted with the initial value that is given here: https://github.com/GoogleCloudPlatform/google-cloud-spanner-hibernate/blob/e17095bdaf584c9f471043b870722944a6ac74d5/google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/BitReversedSequenceStyleGenerator.java#L60
2. The sequence (table) exists (or was just created in step 1): The current value in the table.

So in the example given in the code sample, the sequence (table) will be created with an initial value of 50,000. This value will increase as more and more IDs are generated. The application will read the current value from this table at startup, and when it has exhausted its current range.

google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/BitReversedSequenceStyleGenerator.java
```diff
@@ -0,0 +1,169 @@
...
+ * <p>It is recommended to use a separate table for each generator to prevent a large number of
+ * writes for a single ID generator table. Set the table name to use for a generator with the
+ * SequenceStyleGenerator.SEQUENCE_PARAM parameter (see example below).
+ *
+ * <p>Example usage:
+ *
+ * <pre>{@code
+ * @Id
+ * @GeneratedValue(strategy = GenerationType.SEQUENCE, generator = "customerId")
+ * @GenericGenerator(
```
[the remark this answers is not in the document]
olavloite: True, but the alternative is not available in Hibernate 5, which this change is for, so that will not be possible.

google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/BitReversedSequenceStyleGenerator.java
olavloite: Done

google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/SpannerTableExporter.java
olavloite: Done (although it is a situation that should never happen, so the test is slightly simple/silly)

JoeWang1127 (approved):
LGTM

bharadwaj-aditya (approved):
LGTM

elefeint (approved):
Thank you!

PR #486, opened 2022-11-17
fix: retry transaction for sequence.next_val update

### LINKED ISSUE #485: BitReversedSequence value generation can fail due to concurrent modification exception
The `BitReversedSequenceGenerator` can fail when generating a new sequence value if multiple threads are trying to update the same sequence concurrently. The `AbortedDueToConcurrentModificationException` can (and should) be retried, as Hibernate will execute the update of the sequence value in an isolated transaction.

olavloite:
The read/write transaction that is used to update the sequence value of a table backed sequence can be aborted by Cloud Spanner if multiple entities use the same sequence, and concurrent inserts to those entities require Hibernate to update the sequence value concurrently.

Fixes #485


google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/BitReversedSequenceStyleGenerator.java
```diff
@@ -154,14 +157,45 @@ protected DatabaseStructure buildTableStructure(
...
+        try {
+          Thread.sleep(aborted.getCause().getRetryDelayInMillis());
+        } catch (InterruptedException interruptedException) {
+          Thread.currentThread().interrupt();
+          throw new IdentifierGenerationException("Interrupted while trying to generate a new ID",
+              interruptedException);
+        }
+      }
+    }
     if (id instanceof Long) {
```
JoeWang1127, line 199: Could you add a test case to generate a sequence that is not an instance of `Long`?

google-cloud-spanner-hibernate-dialect/src/test/java/com/google/cloud/spanner/hibernate/BitReversedSequenceStyleGeneratorTests.java
```diff
@@ -44,4 +51,66 @@ protected Serializable generateBaseValue(
...
+          protected Serializable generateBaseValue(
+              SharedSessionContractImplementor session, Object entity) {
+            throw mock(SpannerException.class);
+          }
+        };
+    assertThrows(SpannerException.class, () -> generator.generate(session, customer));
+  }
+
+  @Test
+  public void testGenerateGivesUpEventually() {
```
JoeWang1127, line 97: This test should keep retrying until reach the maximum attempts, the num of retries should also be verified.

google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/BitReversedSequenceStyleGenerator.java
```diff
@@ -154,14 +157,45 @@ protected DatabaseStructure buildTableStructure(
...
+        } else {
+          throw exception;
+        }
+        attempts++;
+        if (attempts == MAX_ATTEMPTS) {
+          throw exception;
+        }
+        try {
+          Thread.sleep(aborted.getCause().getRetryDelayInMillis());
+        } catch (InterruptedException interruptedException) {
```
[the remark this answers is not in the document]
olavloite: Added test

google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/BitReversedSequenceStyleGenerator.java
olavloite: Added a test for a generator that returns an Integer and one that returns a String.

google-cloud-spanner-hibernate-dialect/src/test/java/com/google/cloud/spanner/hibernate/BitReversedSequenceStyleGeneratorTests.java
olavloite: Done

JoeWang1127 (approved):

PR #585, opened 2023-03-20
feat: exclude ranges option for bit-reversed sequence

olavloite:
Support excluding specific ranges from generation by a bit-reversed sequence. This can be used if the bit-reversed sequence is used with a table that contains data that has been imported from a different database, or otherwise has been generated using a different mechanism than the bit-reversed sequence.


google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/BitReversedSequenceStyleGenerator.java
```diff
@@ -139,6 +154,20 @@ public ReplaceInitCommand(String... initCommands) {
...
+   * Configuration property for defining ranges that should be excluded by a bit-reversed sequence
+   * generator.
+   */
+  public static final String EXCLUDE_RANGES_PARAM = "exclude_ranges";
+
+  private QualifiedName sequenceName;
+
+  /**
+   * A set of ranges that should be excluded from generation by this generator. This can be used to
+   * make sure that existing values from a previous system are not re-generated by this generator.
```
arpan14, line 167: Worth mentioning if these ranges are closed or open?

google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/BitReversedSequenceStyleGenerator.java
```diff
@@ -157,8 +187,59 @@ protected DatabaseStructure buildTableStructure(
...
+  }
+
+  @VisibleForTesting
+  void configureExcludedRanges(String sequenceName, Properties params) throws MappingException {
+    this.excludeRanges = parseExcludedRanges(sequenceName, params);
+  }
+
   @VisibleForTesting
-  static final int MAX_ATTEMPTS = 100;
+  static List<Range<Long>> parseExcludedRanges(String sequenceName, Properties params) {
```
arpan14, line 203: Nit: Can we have this method arranged after all the public methods of the class?

google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/BitReversedSequenceStyleGenerator.java
```diff
@@ -168,39 +249,46 @@ protected DatabaseStructure buildTableStructure(
...
+          } catch (InterruptedException interruptedException) {
+            Thread.currentThread().interrupt();
+            throw new IdentifierGenerationException(
+                "Interrupted while trying to generate a new ID", interruptedException);
+          }
         }
       }
+      if (id instanceof Long) {
+        Long reversed = Long.reverse((Long) id);
+        if (excludeRanges.stream().noneMatch(range -> range.contains(reversed))) {
```
arpan14, line 285: Would be good to add info logs when we couldn't generate ID due to excluded ranges? Probably in the `else` block?

arpan14 (approved):

arpan14:
Can we include documentation about this ID generator in the README doc?

google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/BitReversedSequenceStyleGenerator.java
olavloite: Added.

google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/BitReversedSequenceStyleGenerator.java
olavloite: I would prefer not to do that. These methods form a logical block of code together with the `public configure(..)` method. The main reason that it consists of a couple of package-private methods together with the public `configure` method is to make it more testable. Keeping methods that form a single block of logic together improves the readability of that block, as you don't have to jump back and forth a lot in the source code.

Note that the [Java style guide](https://google.github.io/styleguide/javaguide.html#s3.4.2-ordering-class-contents) requires the source code to use some form of logical ordering, and has some strict rules for specific content such as constructors and overloads, but not for the order of `public` and `private` methods.

asthamohta (approved):

PR #694, opened 2023-08-08
chore: add translator for LIKE statements

olavloite:
Adds a translator for LIKE statements, as Cloud Spanner does not support ESCAPE clauses.

### DIFF
--- a/google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/SpannerSqlAstTranslator.java
+++ b/google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/SpannerSqlAstTranslator.java
@@ -1,0 +1,63 @@
+/*
+ * Copyright 2019-2023 Google LLC
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA
+ */
+
+package com.google.cloud.spanner.hibernate;
+
+import org.hibernate.engine.spi.SessionFactoryImplementor;
+import org.hibernate.sql.ast.tree.Statement;
+import org.hibernate.sql.ast.tree.predicate.LikePredicate;
+import org.hibernate.sql.exec.spi.JdbcOperation;
+
+/** We need a translator for the LIKE operator, as Cloud Spanner does not support ESCAPE clauses. */
+public class SpannerSqlAstTranslator<T extends JdbcOperation>
+    extends org.hibernate.dialect.SpannerSqlAstTranslator<T> {
+
+  public SpannerSqlAstTranslator(SessionFactoryImplementor sessionFactory, Statement statement) {
+    super(sessionFactory, statement);
+  }
+
+  @Override
+  public void visitLikePredicate(LikePredicate likePredicate) {
+    // Cloud Spanner does not support ESCAPE clauses.
+    if (likePredicate.isCaseSensitive()) {
+      likePredicate.getMatchExpression().accept(this);
+      if (likePredicate.isNegated()) {
+        appendSql(" not");
+      }
+      appendSql(" like ");
+      likePredicate.getPattern().accept(this);
+    } else {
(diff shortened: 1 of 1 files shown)


google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/SpannerSqlAstTranslator.java
```diff
@@ -0,0 +1,63 @@
...
+    super(sessionFactory, statement);
+  }
+
+  @Override
+  public void visitLikePredicate(LikePredicate likePredicate) {
+    // Cloud Spanner does not support ESCAPE clauses.
+    if (likePredicate.isCaseSensitive()) {
+      likePredicate.getMatchExpression().accept(this);
+      if (likePredicate.isNegated()) {
+        appendSql(" not");
```
arpan14, line 40: Nit: should we append whitespace through `appendSql(WHITESPACE);` ?

google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/SpannerSqlAstTranslator.java
```diff
@@ -0,0 +1,63 @@
...
+
+package com.google.cloud.spanner.hibernate;
+
+import org.hibernate.engine.spi.SessionFactoryImplementor;
+import org.hibernate.sql.ast.tree.Statement;
+import org.hibernate.sql.ast.tree.predicate.LikePredicate;
+import org.hibernate.sql.exec.spi.JdbcOperation;
+
+/** We need a translator for the LIKE operator, as Cloud Spanner does not support ESCAPE clauses. */
+public class SpannerSqlAstTranslator<T extends JdbcOperation>
```
arpan14, line 27: * From the documentation, this class looks very specific to `LIKE` operator. But `SpannerSqlAstTranslator` makes it sound too generic? Should we change this to `SpannerSqlLikeOperatorTranslator` ?
* Or are we saying that `LIKE` operator is the only translator that we need today. And future translators would be added in the same class by overrding new base methods in the future?

google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/SpannerSqlAstTranslator.java
```diff
@@ -0,0 +1,63 @@
...
+    extends org.hibernate.dialect.SpannerSqlAstTranslator<T> {
+
+  public SpannerSqlAstTranslator(SessionFactoryImplementor sessionFactory, Statement statement) {
+    super(sessionFactory, statement);
+  }
+
+  @Override
```
arpan14, line 34: I couldn't find the base method that we are overriding here - https://github.com/hibernate/hibernate-orm/blob/3eb324f714977c8756e21b450fbc33e271fdea6e/hibernate-core/src/main/java/org/hibernate/dialect/SpannerSqlAstTranslator.java . Am I not looking at the right base class?

google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/SpannerSqlAstTranslator.java
olavloite: I don't think that has any additional value:
1. The `WHITESPACE` constant is not environment-specific. It is just `WHITESPACE = ' '` (so a single `char`).
2. It would mean that we would have to do two separate `appendSql(..)` calls, which is less efficient than one.

google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/SpannerSqlAstTranslator.java
olavloite: It's a class that extends from `org.hibernate.dialect.SpannerSqlAstTranslator`, which is defined in the core Hibernate repository (it is part of the `SpannerDialect` that is defined there). That class again is a sub-class of `AbstractSqlAstTranslator`. The reason that the name is so generic, is that the class is intended to contain translators for all SQL operations for a given dialect. Almost all is defined in the generic base class, as most SQL databases use very similar dialects. The only method(s) that we need to override are the ones where we deviate from the standard.

So the reason that is only contains an override for `LIKE` is that this is the only type of expression where Cloud Spanner deviates enough from the standard that we actually need an override. But it could be more in the future.

google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/SpannerSqlAstTranslator.java
olavloite: The method is defined one layer higher up in the class hierarchy here: https://github.com/hibernate/hibernate-orm/blob/0ecd5d8a45f28131edd8dc5a18cfbe97d91f92fc/hibernate-core/src/main/java/org/hibernate/sql/ast/spi/AbstractSqlAstTranslator.java#L7272

arpan14 (approved):

PR #697, opened 2023-08-08
chore: update DDL batch classes

olavloite:


google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/schema/StartBatchDdl.java
```diff
@@ -47,20 +46,21 @@ public boolean appliesToDialect(Dialect dialect) {
 
   @Override
   public boolean beforeTablesOnCreation() {
-    return schemaAction != Action.UPDATE;
+    return schemaAction == Action.CREATE || schemaAction == Action.CREATE_ONLY
+        || schemaAction == Action.UPDATE;
```
arpan14, line 50: Had a question on what this change could mean? Was there a bug previously and this fixes it?

olavloite: Yes, this fixes a bug where this would not be generated in the right position when the value was `UPDATE` or `DROP`. We also need to look into this for the Hibernate 5 dialect, but we currently do not have the test infrastructure in place to do that (it's partly being added in https://github.com/GoogleCloudPlatform/google-cloud-spanner-hibernate/pull/718 and https://github.com/GoogleCloudPlatform/google-cloud-spanner-hibernate/pull/726)

arpan14 (approved):

PR #698, opened 2023-08-08
chore: update SchemaUtils

olavloite:


google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/schema/SchemaUtils.java
```diff
@@ -132,9 +121,7 @@ public static Set<SpannerEntityFieldKey> resolveIdFields(Class<?> entity, Set<Cl
     return ids;
   }
 
-  /**
-   * Returns the field marked with the {@link EmbeddedId} annotation on a class if it exists.
-   */
+  /** Returns the field marked with the {@link EmbeddedId} annotation on a class if it exists. */
```
arpan14, line 124: Is this the right formatting options? I am wondering if this is consistent with how our java client does formatting. It feels a little different. Usually we have a multi line docs for public methods.

arpan14 (approved):

google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/schema/SchemaUtils.java
olavloite: The Hibernate project (unfortunately) uses a different formatter than the client libraries. We should change this so Hibernate uses the same as the client lib, but it's probably better to do that in one go at a later moment, than halfway through this migration.

PR #701, opened 2023-08-08
chore: update types

olavloite:


google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/types/internal/ArrayJavaTypeDescriptor.java
```diff
@@ -55,17 +52,15 @@ public ArrayJavaTypeDescriptor() {
   }
 
   @Override
-  public List<?> fromString(String string) {
+  public List<?> fromString(CharSequence string) {
     throw new UnsupportedOperationException("Creating a Java list from String is not supported.");
```
arpan14, line 56: `Java list from char sequence`

arpan14 (approved):

arpan14 (approved):

google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/types/internal/ArrayJavaTypeDescriptor.java
olavloite: I'm not exactly sure what you mean in this case, but:
1. The method inherits from here: https://github.com/hibernate/hibernate-orm/blob/0ecd5d8a45f28131edd8dc5a18cfbe97d91f92fc/hibernate-core/src/main/java/org/hibernate/type/descriptor/java/BasicJavaType.java#L38. The input argument changed from `String` to `CharSequence` between Hibernate 5 and Hibernate 6.
2. The `List<?>` comes from the fact that the Hibernate dialect has one generic type for all array types, meaning that the `List` could contain a number of different types.

PR #718, opened 2023-08-21
feat!: add support for bit-reversed sequences (server-side)

olavloite:
Adds support for server-side bit-reversed sequences and a custom ID generator that supports the combination of:
1. Server-side bit-reversed sequences.
2. `increment_size > 1`, which again means that the generator does not need a single round-trip to the database for each insert.

Recommended configuration to use a bit-reversed sequence is:

```java
  @Id
  @GeneratedValue(
      strategy = GenerationType.SEQUENCE,
      generator = "my_entity_id_generator"
  )
  @GenericGenerator(
      name = "my_entity_id_generator",
      strategy = "com.google.cloud.spanner.hibernate.PooledBitReversedSequenceStyleGenerator",
      parameters = {
          @Parameter(name = "sequence_name", value = "my_sequence"),
          @Parameter(name = "increment_size", value = "200"),
          @Parameter(name = "initial_value", value = "5000"),
          @Parameter(name = "exclude_range", value = "[1,1000]")
      }
  )
  private Long id;
```

### Breaking Change
This changes the return value for `SpannerDialect#supportsSequences()` from `false` to `true`. This means that any entity that has been set up with a no-argument `@GeneratedValue` will change from using a table-backed sequence to a real sequence. You can prevent this change from affecting your application by adding an explicit strategy to your `@GeneratedValue` annotation.

Change this:

```java
  @Id
  @GeneratedValue
  private long id;
```

Into this:

```java
  @Id
  @GeneratedValue(strategy = GenerationType.TABLE)
  private long id;
```


olavloite:
@dttran FYI

google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/BitReversedSequenceStyleGenerator.java
```diff
@@ -203,46 +213,6 @@ void configureExcludedRanges(String sequenceName, Properties params) throws Mapp
     this.excludeRanges = parseExcludedRanges(sequenceName, params);
   }
 
-  @VisibleForTesting
```
olavloite, line 206: This method is moved to the `EnhancedBitReversedSequenceStyleGenerator`, as both generators can share this code. It makes more sense to have it in the non-deprecated class than in this one.

PR #940, opened 2024-02-27
feat: query hints

olavloite:
Adds support for query hints in the Spanner Hibernate dialect.


google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/SpannerDialect.java
```diff
@@ -256,4 +259,48 @@ public SqmMultiTableInsertStrategy getFallbackSqmInsertStrategy(
...
+
+  @Override
+  public String addSqlHintOrComment(
+      String sql,
+      QueryOptions queryOptions,
+      boolean commentsEnabled) {
+    if (hasCommentHint(queryOptions)) {
+      sql = applyHint(sql, queryOptions.getComment());
+    }
+    if (queryOptions.getDatabaseHints() != null && !queryOptions.getDatabaseHints().isEmpty()) {
```
ankiaga, line 271: Cant we use CollectionUtils.isNotEmpty() from apache commons?

olavloite: Apache Commons is not a dependency in this library, so no.

google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/hints/ReplaceQueryPartsHint.java
```diff
@@ -0,0 +1,335 @@
...
+    return new ReplaceQueryPartsHint(
+        ImmutableList.of(new Replacement(regex, replacement, replaceMode)));
+  }
+
+  protected ReplaceQueryPartsHint(ImmutableList<Replacement> replacements) {
+    this.replacements = Preconditions.checkNotNull(replacements);
+  }
+
+  /**
+   * Creates a hint that will replace all occurrences of the keys in the given map with the
```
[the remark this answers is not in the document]
olavloite: Good point, I changed the data structure halfway through the development, but did not update this comment.

google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/SpannerDialect.java
```diff
@@ -256,4 +259,48 @@ public SqmMultiTableInsertStrategy getFallbackSqmInsertStrategy(
...
+      sql = applyQueryHints(sql, queryOptions);
+    }
+    return super.addSqlHintOrComment(sql, queryOptions, commentsEnabled);
+  }
+
+  private String applyHint(String sql, String hint) {
```
ankiaga, line 277: This and other below functions could be static

google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/SpannerDialect.java
```diff
@@ -256,4 +259,48 @@ public SqmMultiTableInsertStrategy getFallbackSqmInsertStrategy(
...
+    try {
+      return ReplaceQueryPartsHint.fromComment(hint).replace(sql);
+    } catch (Throwable ignore) {
+      // Just ignore and continue with the query normally.
```
ankiaga, line 281: Why? This way user would not know about the issue. Should we atleast log the error?

google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/hints/Hints.java
```diff
@@ -0,0 +1,515 @@
...
+
+  /**
+   * Creates a hint that adds @{USE_ADDITIONAL_PARALLELISM=value} to the statement.
+   *
+   * @param value the hint value
+   * @return a hint that can be added as a comment or query hint to a Hibernate statement
+   */
+  public static ReplaceQueryPartsHint useAdditionalParallelism(boolean value) {
+    return ReplaceQueryPartsHint.of(
+        "(?i)(?:^|\\s)(select|insert|update|delete)",
```
ankiaga, line 124: Can we define this string as a constant?
Also can we define a private function which takes 2 args ("USE_ADDITIONAL_PARALLELISM" and value) so that every other function can use that?

google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/SpannerDialect.java
olavloite: No, they're lists, so that won't work.

google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/SpannerDialect.java
olavloite: Done

google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/SpannerDialect.java
olavloite: The reason that we ignore it here is that we don't actually know that it is a hint. All we know is that:
1. Something in the query generation chain has added a  comment to the query.
2. That comment contains some substrings that we expect when it is a hint.
3. BUT: It could in theory also be a coincidence that it contains those keywords, and just a comment that should not be treated as a hint. (We could make the check for substrings that must be in the comment for it to be considered a hint even more strict, but that would change the behavior; hints that for example contain a typo would then not pass that stricter check, and also not cause an error)

I've added a log statement for it.

google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/hints/Hints.java
olavloite: Done

google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/SpannerDialect.java
ankiaga: Oops!

ankiaga (approved):

PR #1074, opened 2024-05-17
ci: only test spring samples when using JDK 17 or later

burkedavison:
These Spring samples use Spring Boot 3.x to be compatible with Hibernate 6, but Spring Boot 3.x is only compatible with JDK 17 or later. Spring Boot 2.x is not compatible with Hibernate 6.


burkedavison:
Cloud Build tests only set up for nightly tests off main. /gcbrun does not work on this repo right now.

olavloite (approved):

PR #1180, opened 2024-07-09
feat: support float32 type

arawind:
Adds support for `FLOAT32` to the Spanner Hibernate dialect. Note that Hibernate by default maps `java.lang.float` fields to the JDBC type `FLOAT`, which is the equivalent of `DOUBLE`, meaning that these fields by default use `FLOAT64` on Spanner.

Add `@JdbcTypeCode(Types.REAL)` to your field declaration to ensure that a `FLOAT32` column is used.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -133,7 +133,7 @@ sequence in the database.
 
 The `increment_size` for this pooled generator can not exceed 200.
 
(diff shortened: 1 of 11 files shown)


arawind:
@olavloite can you please review these changes?

olavloite (approved):

PR #1523, opened 2025-02-28
feat: add support for identity columns

olavloite:
Adds support for identity columns for primary key generation. This type of primary key is generally not recommended for use with Hibernate, as they do not support batching.


hengfengli (approved):
LGTM. Thanks a lot for adding this support!

PR #1675, opened 2025-06-04
chore: upgrade undertow-core

jinseopkim0:
b/416726020

### DIFF
--- a/google-cloud-spanner-hibernate-samples/microprofile-jpa-sample/pom.xml
+++ b/google-cloud-spanner-hibernate-samples/microprofile-jpa-sample/pom.xml
@@ -57,6 +57,12 @@
       <version>4.1.0</version>
     </dependency>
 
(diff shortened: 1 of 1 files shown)


google-cloud-spanner-hibernate-samples/microprofile-jpa-sample/pom.xml
```diff
@@ -57,6 +57,12 @@
       <version>4.1.0</version>
     </dependency>
 
+    <dependency>
+        <groupId>io.undertow</groupId>
+        <artifactId>undertow-core</artifactId>
+        <version>2.3.11.Final</version>
```
JoeWang1127, line 63: Why do we need to add this dependnecy, is this because Maven can pull the latest version of `undertow-core`, rather than the version imported from `thorntail`?

PR #1785, opened 2025-08-11
docs: update readme for usage of @GeneratedValue and select for update

rayudu3745:


README.adoc
```diff
@@ -118,38 +118,40 @@ The `@Type(type="uuid-char")` annotation specifies that this UUID value will be
...
 __NOTE__: Read to the end of this section to see the recommended way to set up `@GeneratedValue`.
 
-Hibernate's `@GeneratedValue` annotation for numeric fields is supported, and will by default use a
-positive bit-reversed sequence. A bit-reversed sequence internally uses a monotonically increasing
-counter  that is reversed before being returned to Hibernate. This means that the identifiers that
-are generated are in the form bitReversePositive(1), bitReversePositive(2), ...:
-[source, java]
+❌ Not recommended
+
+`GenerationType.AUTO`, `GenerationType.TABLE`, `GenerationType.SEQUENCE` - Not supported by the Spanner dialect.
```
olavloite, line 125: nit:

```suggestion
`GenerationType.AUTO`, `GenerationType.TABLE`, `GenerationType.SEQUENCE`: These are not supported by the Spanner dialect.
```

Note for the longer term (so not for this PR):
1. We should support AUTO. That should lead to the 'least-bad' auto-generation strategy that we can support, which I believe with our current feature set would be IDENTITY.
2. We should support TABLE if possible. I believe that the current limitation is that Hibernate generates an initialization command that tries to insert a row without specifying the column name. We should try to either override that in our dialect, or change this in Hibernate itself.

README.adoc
```diff
@@ -118,38 +118,40 @@ The `@Type(type="uuid-char")` annotation specifies that this UUID value will be
...
+
+`GenerationType.IDENTITY`
+-   See https://cloud.google.com/spanner/docs/primary-key-default-value#identity-columns[Identity columns].
+IDs are generated at insert time, which prevents Hibernate from batching inserts efficiently.
```
olavloite, line 129:
```suggestion
It is not recommended to use the `IDENTITY` generation strategy, as it prevents Hibernate from using JDBC batching for inserts. See https://docs.jboss.org/hibernate/orm/current/userguide/html_single/Hibernate_User_Guide.html#batch-session-batch for more information.
```

README.adoc
```diff
@@ -118,38 +118,40 @@ The `@Type(type="uuid-char")` annotation specifies that this UUID value will be
...
   public Long id;
 }
 ----
 
-Bit-reversed sequences do not support an increment size larger than 1. This means that entities
-that use this style of identifiers by default require a round-trip to the database for each entity
-that is inserted. The `PooledBitReversedSequenceStyleGenerator` provided in this repository fixes
-this problem by using the configured `increment_size` to generate a query that fetches multiple
-identifier values from the sequence in one query, instead of setting an `increment_size` on the
-sequence in the database.
```
olavloite, line 146: Why do we want to remove this?

README.adoc
```diff
@@ -118,38 +118,40 @@ The `@Type(type="uuid-char")` annotation specifies that this UUID value will be
...
+
+`GenerationType.AUTO`, `GenerationType.TABLE`, `GenerationType.SEQUENCE` - Not supported by the Spanner dialect.
+
+`GenerationType.IDENTITY`
+-   See https://cloud.google.com/spanner/docs/primary-key-default-value#identity-columns[Identity columns].
+IDs are generated at insert time, which prevents Hibernate from batching inserts efficiently.
+
+
+`@GeneratedValue` +  `@SequenceGenerator` with `allocationSize = 1`
+Uses https://cloud.google.com/spanner/docs/primary-key-default-value#bit-reversed-sequence[Bit-reversed sequences] by spanner. Batching is inhibited because each insert requires a round‑trip to database
```
olavloite, line 133:
```suggestion
This strategy uses a https://cloud.google.com/spanner/docs/primary-key-default-value#bit-reversed-sequence[bit-reversed sequence] with an allocation size of 1. This requires Hibernate to fetch a new
sequence value for each insert, causing each insert to cause two round-trips to Spanner.
```

README.adoc
```diff
@@ -683,8 +684,7 @@ mutations for Cloud Spanner. The dialect will only generate DML statements.
 
 === Locking
 
-Cloud Spanner does not support explicit locking clauses like `SELECT ... FOR UPDATE`. Setting an explicit lock level in
-Hibernate is therefore not supported. The following is for example not supported:
+Cloud Spanner supports exclusive row locks using https://cloud.google.com/spanner/docs/use-select-for-update[`SELECT ... FOR UPDATE`]
```
olavloite, line 687:
```suggestion
Cloud Spanner supports explicit row locks using https://cloud.google.com/spanner/docs/use-select-for-update[`SELECT ... FOR UPDATE`]
```

These locks are not 'exclusive', as they do allow reads to continue to happen on these rows.

olavloite (approved):

PR #2020, opened 2026-08-04
feat: upgrade Hibernate ORM to version 7.4.5.Final

rayudu3745:
- Bump <hibernate.version> to 7.4.5.Final in pom.xml
- Remove 620 lines of duplicate/redundant overrides from SpannerDialect, SpannerIdentityColumnSupport, and SpannerSqlAstTranslator in favor of upstream org.hibernate.dialect.SpannerDialect
- Override getCreateIndexString(boolean) to use standard CREATE UNIQUE INDEX syntax and initDefaultProperties() to set PREFERRED_POOLED_OPTIMIZER to pooled-lo
- Update expected identity column DDL string in SchemaGenerationMockServerTest to match upstream Hibernate 7.4.5's lowercase generated by default as identity format
- Replace hardcoded sequence metadata query in AbstractSchemaGenerationMockServerTest with dynamic SpannerDialect().getQuerySequencesString().trim() statement
- Restructure column DDL generation in SpannerTableStatements to cleanly separate identity columns from nullable/default columns
- Set created_at and updated_at timestamps to Timestamp.now() in SingerService raw Spanner SDK insert mutations because Hibernate 7.4 generates these columns as NOT NULL by default in DDL schema export
- Update SampleApplicationMockServerTest to support aliased SQL update queries and dynamically trim sequence query statements


google-cloud-spanner-hibernate-dialect/src/test/java/com/google/cloud/spanner/hibernate/SpannerDialectTests.java
```diff
@@ -198,7 +198,7 @@ public void getDropStringTest() {
 
   @Test
   public void getCreateTableStringTest() {
-    assertThat(this.spannerDialect.getCreateTableString()).isEqualTo("create table if not exists");
+    assertThat(this.spannerDialect.getCreateTableString()).isEqualTo("create table");
```
sakthivelmanii, line 201: did we change the builtin dialect?

rayudu3745: yes we are the BaseDialect one, no much use for adding if not exists here 
some testcases are failing for if we added if not exists

olavloite (approved):

